PR #1, opened 2024-03-18
Fix nix setup after move to separate repo

infinisil:
After the move into this new repository (https://github.com/NixOS/nixpkgs/issues/286559), some changes are necessary to make it work, since there's some Nix code that relied on `../../..` to be Nixpkgs.

- Adds a pinned Nixpkgs using [npins](https://github.com/andir/npins)
- Move `default.nix` to `package.nix` and move things that are needed outside of it to `default.nix`
- Fix `shell.nix` to use the `shell` attribute from `default.nix`.

This only fixes the Nix setup, I'll create other PRs for e.g. the Readme and co.


philiptaron (approved):
I'll make a flake since that's what I use, but it'll use `default.nix`.

infinisil:
@philiptaron Sounds good, you could base it on https://github.com/NixOS/nixfmt/blob/master/flake.nix

PR #10, opened 2024-03-18
Add rust-analyzer to the dev env

infinisil:
Apparently I got it from my local environment up to now :sweat_smile: 

Discovered in a call with @philiptaron and @willbush


philiptaron (approved):
yes, you did

PR #12, opened 2024-03-19
Set up CI to automatically update Nixpkgs

infinisil:
Yet untested


infinisil:
I guess we'll see whether it works after merging

.github/workflows/update.yml
```diff
@@ -0,0 +1,26 @@
+name: Update Nixpkgs pin
+on:
+  workflow_dispatch: # ALlows triggering manually
```
philiptaron, line 3:
```suggestion
  workflow_dispatch: # Allows triggering manually
```

.github/workflows/update.yml
```diff
@@ -0,0 +1,26 @@
...
+
+      - uses: cachix/install-nix-action@v26
+
+      - name: update
+        run: |
+          nix-build -A pkgs.npins
+          result/bin/npins update nixpkgs
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v6
```
philiptaron, line 21: I would have thought there was some other PR-opening action, but this one by some guy is the most used. Odd.

philiptaron (approved):

philiptaron (approved):

.github/workflows/update.yml
infinisil: I guess it's fairly standard to have third-party actions

philiptaron: I expected it to have some sort of ... github imprimatur.

philiptaron:
[It doesn't work.](https://github.com/NixOS/nixpkgs-check-by-name/pull/14)

PR #13, opened 2024-03-19
Add editorconfig

infinisil:
https://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+indent_style = space
+max_line_length = 100
+
+[*.rs]
+indent_size = 4
+
+[*.nix]
+indent_size = 2


infinisil:
We could also add CI to check that, but I don't think it's that important. I think it's just good practice for all repos to have one of those :)

.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+indent_style = space
+
+[*.rs]
+indent_size = 4
```
philiptaron, line 11: I like adding `max_line_length = 100` (maybe for all)

I would also like to use the nixfmt-rfc-style nix formatter!

https://github.com/rust-lang/rust/blob/master/.editorconfig

philiptaron (commented):
What do you think?

.editorconfig
infinisil: Let's do Nix formatting with another PR, but the max line length sounds good, adding it for `[*]` :)

philiptaron (approved):

PR #16, opened 2024-03-21
Automated releases

infinisil:
This sets up automated releases, mainly so that we can distribute a pre-built version of the tool, allowing Nixpkgs' CI to fetch it without building.

I used the release script to generate the first release, which also contains some info about how to use the artifact: https://github.com/NixOS/nixpkgs-check-by-name/releases/tag/0.1.0

See [this Matrix message](https://matrix.to/#/!djTaTBQyWEPRQxrPTb:nixos.org/$1v7Jnqa3ixC_FHPOILFkqrde3kDgjvihXZUZ_Pz6ErA?via=nixos.org&via=matrix.org&via=nixos.dev) for historical context.

After this and #15 is merged, we could also extend the automated PRs to regularly update the version, such that we don't forget to do new releases.

PR #18, opened 2024-03-21
Set `GH_TOKEN` to fix automated releases

infinisil:
Hopefully..

See https://github.com/NixOS/nixpkgs-check-by-name/actions/runs/8381658441/job/22953710054#step:4:575 for the failure and #16 for where this was introduced

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,3 +14,5 @@ jobs:
 
       - name: release
         run: scripts/release.sh
+        env:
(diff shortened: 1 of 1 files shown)


philiptaron (approved):

infinisil:
Nice it works: https://github.com/NixOS/nixpkgs-check-by-name/actions/runs/8382741125/job/22957094421

Well, we'll see if it _really_ works when we increase the version number the next time :P

PR #19, opened 2024-03-21
Don't commit PR body for automated updates

infinisil:
Fixes https://github.com/NixOS/nixpkgs-check-by-name/pull/17#discussion_r1534779063, the PR should fix itself automatically once triggered again

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -9,12 +9,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        path: repo
 
       - uses: cachix/install-nix-action@v26
 
       - name: update
         run: |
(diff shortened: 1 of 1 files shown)


philiptaron (approved):
Let's give it a go.

infinisil:
> The workflow is not valid. .github/workflows/update.yml (Line: 12, Col: 9): Unexpected value 'path'

Huh (from [here](https://github.com/NixOS/nixpkgs-check-by-name/actions/runs/8382774071))

.github/workflows/update.yml
```diff
@@ -9,12 +9,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        path: repo
```
infinisil, line 12: Oh it should be

```
with:
  path: repo
```

PR #21, opened 2024-03-21
Fix updates (another try)

infinisil:
Addresses the previous problem (https://github.com/NixOS/nixpkgs-check-by-name/pull/20#issuecomment-2013998832) with the second commit.

Also using this opportunity to improve how we define what CI should build and make it build the automated update script (but not running it), see the first commit for that.

This makes sure that when we make changes to the update script (like in this PR), it at least gets at least built (though not ran).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,4 +13,4 @@ jobs:
       - uses: cachix/install-nix-action@v26
 
       - name: build
-        run: nix-build -A build -A shell -A checks
+        run: nix-build -A ci
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -17,7 +17,7 @@ jobs:
       - name: update
         run: |
           nix-build repo -A autoPrUpdate
-          result/bin/auto-pr-update > body
+          result/bin/auto-pr-update repo > body
 
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v6
--- a/default.nix
+++ b/default.nix
@@ -56,19 +56,22 @@ let
       ];
       text =
         let
-          commands = [
-            "npins update"
-            "cargo update"
-          ];
+          commands = {
+            "npins changes" = ''
+              npins update --directory "$REPO_ROOT/npins"'';
+            "cargo changes" = ''
+              cargo update --manifest-path "$REPO_ROOT/Cargo.toml"'';
+          };
         in
         ''
+          REPO_ROOT=$1
           echo "Run automated updates"
         ''
-        + pkgs.lib.concatMapStrings (command: ''
-          echo -e '<details><summary>${command}</summary>\n\n```'
+        + pkgs.lib.concatStrings (pkgs.lib.mapAttrsToList (title: command: ''
+          echo -e '<details><summary>${title}</summary>\n\n```'
           ${command} 2>&1
           echo -e '```\n</details>'
-        '') commands;
+        '') commands);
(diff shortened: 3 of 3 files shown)


default.nix
```diff
@@ -30,63 +30,71 @@ let
...
-    env.NIX_CHECK_BY_NAME_EXPR_PATH = toString runtimeExprPath;
-    env.NIX_PATH = "test-nixpkgs=${toString testNixpkgsPath}:test-nixpkgs/lib=${toString nixpkgsLibPath}";
-    env.RUST_SRC_PATH = "${pkgs.rustPlatform.rustLibSrc}";
-    inputsFrom = [ build ];
-    nativeBuildInputs = with pkgs; [
-      npins
-      rust-analyzer
-    ];
+    # Tests the tool on the pinned Nixpkgs tree, this is a good sanity check
+    nixpkgsCheck = pkgs.runCommand "test-nixpkgs-check-by-name" {
```
philiptaron, line 78: We're going to need an escape hatch for this.

.github/workflows/main.yml
```diff
@@ -13,4 +13,4 @@ jobs:
       - uses: cachix/install-nix-action@v26
 
       - name: build
-        run: nix-build -A build -A shell -A checks
+        run: nix-build -A ci
```
philiptaron, line 16: Nice, I like this.

.github/workflows/update.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: update
         run: |
           nix-build repo -A autoPrUpdate
-          result/bin/auto-pr-update > body
+          result/bin/auto-pr-update repo > body
```
philiptaron, line 20: Do we maybe want to add `body` to gitignore?

default.nix
```diff
@@ -30,63 +30,71 @@ let
...
-        echo "Run automated updates"
-      ''
-      + pkgs.lib.concatMapStrings (command: ''
-        echo -e '<details><summary>${command}</summary>\n\n```'
-        ${command} 2>&1
-        echo -e '```\n</details>'
-      '') commands;
-  };
+in
+results.build // results // {
```
philiptaron, line 92: why do we need `results.build` to be first?

infinisil: Shouldn't matter, they need to be non-overlapping. In the future I'd like to use `lib.derivations.lazyDerivation` (to avoid `results.build` having so many attributes) and `lib.attrsets.unionOfDisjoint` (to ensure they don't overlap) to make this nicer.

.github/workflows/update.yml
infinisil: Feels overkill, this should be fixed now :)

default.nix
infinisil: How do you mean?

infinisil:
Btw best reviewed with whitespace hidden

default.nix
philiptaron: I mean, as we add things that cause rightfully-detected errors in nixpkgs, we'll need to allow those errors to pass CI.

philiptaron (approved):

default.nix
infinisil: This sanity check doesn't run any ratchet checks, because it sets base Nixpkgs = main Nixpkgs :)

infinisil:
We're almost there.. https://github.com/NixOS/nixpkgs-check-by-name/pull/22

PR #24, opened 2024-03-22
Refactor Error Types

willbush:
Work related to: https://github.com/NixOS/nixpkgs-check-by-name/issues/2

This pull request does not change the single-line vs multi-line style or any of the error messages. The goal here is to reduce code duplication.

A lot of the NixpkgsProblem enum variants had common fields more or less due to the context of where the error originated.

I grouped these similar problems into structs which contains the common fields and a `kind` enum to differentiate between the different types of problems.

### DIFF
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -213,8 +213,6 @@ fn by_name(
     use ratchet::RatchetState::*;
     use ByNameAttribute::*;
 
-    let relative_package_file = structure::relative_file_for_package(attribute_name);
-
     // At this point we know that `pkgs/by-name/fo/foo/package.nix` has to exists.
     // This match decides whether the attribute `foo` is defined accordingly
     // and whether a legacy manual definition could be removed
@@ -224,7 +222,6 @@ fn by_name(
             // This indicates a bug in the `pkgs/by-name` overlay, because it's supposed to
             // automatically defined attributes in `pkgs/by-name`
             NixpkgsProblem::UndefinedAttr {
-                relative_package_file: relative_package_file.to_owned(),
                 package_name: attribute_name.to_owned(),
             }
             .into()
@@ -242,7 +239,6 @@ fn by_name(
             // We can't know whether the attribute is automatically or manually defined for sure,
             // and while we could check the location, the error seems clear enough as is.
             NixpkgsProblem::NonDerivation {
-                relative_package_file: relative_package_file.to_owned(),
                 package_name: attribute_name.to_owned(),
             }
             .into()
@@ -262,7 +258,6 @@ fn by_name(
                 Success(())
             } else {
                 NixpkgsProblem::NonDerivation {
-                    relative_package_file: relative_package_file.to_owned(),
                     package_name: attribute_name.to_owned(),
                 }
                 .into()
@@ -312,7 +307,6 @@ fn by_name(
 
                         by_name_override(
                             attribute_name,
-                            relative_package_file,
                             is_semantic_call_package,
                             optional_syntactic_call_package,
                             definition,
@@ -350,7 +344,6 @@ fn by_name(
 /// all-packages.nix
 fn by_name_override(
     attribute_name: &str,
-    expected_package_file: RelativePathBuf,
     is_semantic_call_package: bool,
     optional_syntactic_call_package: Option<CallPackageArgumentInfo>,
     definition: String,
@@ -381,6 +374,8 @@ fn by_name_override(
         // Something like `<attr> = pkgs.callPackage ...`
         (true, Some(syntactic_call_package)) => {
             if let Some(actual_package_file) = syntactic_call_package.relative_path {
+                let expected_package_file = structure::relative_file_for_package(attribute_name);
+
                 if actual_package_file != expected_package_file {
                     // Wrong path
                     NixpkgsProblem::WrongCallPackagePath {
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -352,7 +352,7 @@ fn by_name_override(
     location: Location,
     relative_location_file: RelativePathBuf,
 ) -> validation::Validation<ratchet::RatchetState<ratchet::ManualDefinition>> {
-    let to_problem = |kind| -> NixpkgsProblem {
+    let to_problem = |kind| {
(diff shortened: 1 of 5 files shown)


willbush:
I'm not able to add reviewers @infinisil @philiptaron :)

src/nixpkgs_problem.rs
```diff
@@ -11,390 +11,410 @@ use std::fmt;
...
+
+#[derive(Clone)]
+pub enum RatchetErrorKind {
+    MovedOutOfByName,
+    MovedOutOfByNameEmptyArg,
+    NewPackageNotUsingByName,
+    NewPackageNotUsingByNameEmptyArg,
 }
 
 impl fmt::Display for NixpkgsProblem {
```
philiptaron, line 139: These all want to be split up as well.

src/ratchet.rs
```diff
@@ -153,32 +153,16 @@ impl ToNixpkgsProblem for UsesByName {
...
-                file: file.to_owned(),
-            }
-        } else {
-            NixpkgsProblem::NewPackageNotUsingByNameNonEmptyArg {
-                package_name: name.to_owned(),
-                call_package_path: to.relative_path.clone(),
-                file: file.to_owned(),
-            }
-        }
+        NixpkgsProblem::Ratchet(RatchetError {
```
philiptaron, line 156: This is nicer.

src/references.rs
```diff
@@ -125,46 +128,35 @@ fn check_nix_file(
 
     Ok(validation::sequence_(
         nix_file.syntax_root.syntax().descendants().map(|node| {
-            let text = node.text().to_string();
-            let line = nix_file.line_index.line(node.text_range().start().into());
-
             // We're only interested in Path expressions
-            let Some(path) = rnix::ast::Path::cast(node) else {
+            let Some(path) = rnix::ast::Path::cast(node.clone()) else {
```
philiptaron, line 132: This clone makes me sad. I think it's only needed because of the to_problem closure. Could you tell me why it's there now?

philiptaron (commented):
Some initial thoughts. I like the direction.

infinisil (changes requested):
Reviewed together in today's office hour with @willbush and @philiptaron

Got some minor suggestions, overall looking good.

One other thing is to add simple comments to the new types.

src/nixpkgs_problem.rs
```diff
@@ -11,390 +11,410 @@ use std::fmt;
...
+}
+
+#[derive(Clone)]
+pub struct ShardError {
+    pub shard_name: String,
+    pub kind: ShardErrorKind,
+}
+
+#[derive(Clone)]
+pub enum ShardErrorKind {
```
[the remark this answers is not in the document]
infinisil: https://github.com/NixOS/nixpkgs-check-by-name/pull/24#discussion_r1538161160

src/references.rs
willbush: It was only needed because I moved `line` and `text` closer to where they were used without thinking much about how big the node object might be. I changed it back to remove the clone.

philiptaron: # 🔥 
Love that.

src/eval.rs
```diff
@@ -213,7 +215,12 @@ fn by_name(
     use ratchet::RatchetState::*;
     use ByNameAttribute::*;
 
-    let relative_package_file = structure::relative_file_for_package(attribute_name);
+    let to_problem = |kind| {
```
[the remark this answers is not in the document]
willbush: I was able to change most of the `to_problem` into `to_validation` here:

https://github.com/NixOS/nixpkgs-check-by-name/pull/24/commits/2ed5984984b956a9562848f3117427f4ca2aec81

In some cases the `into` was resolving into different types. In one case (https://github.com/NixOS/nixpkgs-check-by-name/pull/24/files#diff-84c46c1047bea02643d764007dc4ae2cb6e93767ba92faa9583659cb88b49bf4R263) I could resolve that by using `map`

However, in this case, it's not so simple so I just left it as to_problem:

https://github.com/NixOS/nixpkgs-check-by-name/pull/24/files#diff-84c46c1047bea02643d764007dc4ae2cb6e93767ba92faa9583659cb88b49bf4R382

src/nixpkgs_problem.rs
```diff
@@ -143,6 +124,22 @@ pub enum NixpkgsProblem {
     },
 }
 
+#[derive(Clone)]
+pub struct RatchetError {
```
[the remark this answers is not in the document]
willbush: Ok got this fixed and added some comments.

src/eval.rs
```diff
@@ -213,7 +214,13 @@ fn by_name(
     use ratchet::RatchetState::*;
     use ByNameAttribute::*;
 
-    let relative_package_file = structure::relative_file_for_package(attribute_name);
+    let to_validation = |kind| -> validation::Validation<RatchetState<ManualDefinition>> {
```
philiptaron, line 217: I expected this to be the same return type as the function it's embedded in: `validation::Result<ratchet::Package>`. For my curiosity, why is it different?

On a naming front, I would name this `into_result`, if the transformation to have it return a `Result` is possible. It's short and reads nicely at callsites. `into` is the word for consuming what's passed in, which this helper does.

philiptaron (approved):
Some curiosities.

src/eval.rs
willbush: The `to_validation` closure has the same return type as `manual_definition_result`. Here it is with the explicit type:

```rust
    let manual_definition_result: validation::Validation<RatchetState<ManualDefinition>> = match by_name_attribute {
        // ...
    }
    Ok(
        manual_definition_result.map(|manual_definition| ratchet::Package {
            manual_definition,
            uses_by_name: Tight,
        }),
    )
```

At the end of the `by_name` function is where it's wrapped with `OK()` to become a `Result` and `manual_definition_result` is mapped into a `ratchet::Package`. There's a couple of places within the above match expression where an IO error can occur and early out (e.g. `let nix_file = nix_file_store.get(&location.file)?;`) and return a Result type early.

philiptaron: Thanks.

PR #28, opened 2024-03-26
Include clippy lints for tests

willbush:
Randomly noticed some warnings in `main.rs` since [I improved my rust-analyzer integration in Emacs](https://github.com/willbush/system/blob/6fb4c108ab0a471430aac2ff5c5d925bb31292b7/configs/emacs/default/modules/init-prog-tools.el#L394). Looked into why and apparently `cargo clippy` doesn't include tests by default:

https://github.com/rust-lang/rust-clippy/issues/1436#issuecomment-462059561
https://github.com/rust-lang/rust-clippy/issues/10690

We could pass `--all-targets` or `--tests` to fix the issue. I liked the former because its more obvious that lints are checked for more than *just* tests.

### DIFF
--- a/package.nix
+++ b/package.nix
@@ -39,7 +39,8 @@ rustPlatform.buildRustPackage {
   preCheck = initNix;
   postCheck = ''
     cargo fmt --check
-    cargo clippy -- -D warnings
+    # --tests or --all-targets include tests for linting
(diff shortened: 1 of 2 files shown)


infinisil (approved):
Oh nice, thanks!

PR #33, opened 2024-04-04
Consistent relative paths to expression file

willbush:
### 🚀 Catalyst

https://github.com/NixOS/nixpkgs-check-by-name/issues/2#issuecomment-2018895221

### 📃 Description

PR has changes needed to output error messages with paths relative to expression
file instead of nixpkgs root.

- `tests: move all-packages.nix to pkgs/top-level`

  In this commit I move all the `tests/*/all-packages.nix` to `tests/*/pkgs/top-level/all-packages.nix` with changing any code (breaking a lot of tests).

  The reason for this commit is to make it more obvious in the `expected` files results where changes where needed to get `./pkgs/by-name` to  become `./../by-name`

- `Fix relative paths in all-packages.nix files and expected outputs`

  This commit just gets tests passing again before I start modifying other code.
- `Update errors paths to be relative to the expression file`

  Most changes are here to use relative paths to expression file in all cases.


willbush:
Annoying. I thought moving files in one commit without modifying meant git would detect it as a move. Guess I should have used `git mv` explicitly. Looking into how to fix it up.

infinisil:
If you look at the diff commit-by-commit it works, so that seems fine to me :)

PR #39, opened 2024-04-10
Add dependabot configuration to update GH actions

willbush:
@infinisil I recall you saying something about feeling dependabot is too Github specific. However, I'm not aware of an easier way to keep Github actions up-to-date. Speaking of not being GH specific, worth considering auto-mirroring this repo to another place just for backup? I thought about doing the same for my personal repos, but never got around to it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "weekly"


infinisil:
Hmm, would it be possible to combine the dependabot updates with the other weekly updates? I'd rather only have one automated PR per week

willbush:
I think since we're using major version numbers, in practice we wont see updates that often. I didn't see cron support. Could make it monthly?

 I'm not aware of a way to combine updates with the release action.

philiptaron (approved):
I think this is a super low rate of change. As a result, I'm fine with it.

infinisil:
I've been working on this: https://github.com/NixOS/nixpkgs-check-by-name/pull/40 :D

PR #40, opened 2024-04-10
Automatic GitHub action updates

infinisil:
This is an alternative to #39. It runs the [dependabot CLI](https://github.com/dependabot/cli) locally in the existing weekly update CI action (along with some jq magic). <s>Yet untested in CI, but </s> It also works locally, which is really nice!

Best reviewed commit-by-commit, there's some Nix changes in the first 2 commits.

---

This work is sponsored by [Antithesis](https://antithesis.com/) :sparkles:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,7 +5,7 @@ on:
       - main
 
 jobs:
-  check:
+  build:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
@@ -14,3 +14,23 @@ jobs:
 
       - name: build
         run: nix-build -A ci
+
+  test-update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: cachix/install-nix-action@v26
+
+      - name: test update script
+        run: |
+          nix-build -A autoPrUpdate
+          {
+            result/bin/auto-pr-update .
+            echo ""
+            echo '```diff'
+            git diff
+            echo '```'
+          } > $GITHUB_STEP_SUMMARY
+        env:
+          GH_TOKEN: ${{ github.token }}
--- a/default.nix
+++ b/default.nix
@@ -53,31 +53,52 @@ let
     };
 
     # Run regularly by CI and turned into a PR
-    autoPrUpdate = pkgs.writeShellApplication {
-      name = "auto-pr-update";
-      runtimeInputs = with pkgs; [
-        npins
-        cargo
-      ];
-      text =
-        let
-          commands = {
-            "npins changes" = ''
-              npins update --directory "$REPO_ROOT/npins"'';
-            "cargo changes" = ''
-              cargo update --manifest-path "$REPO_ROOT/Cargo.toml"'';
+    autoPrUpdate =
+      let
+        updateScripts = {
+          npins = pkgs.writeShellApplication {
+            name = "update-npins";
+            runtimeInputs = with pkgs; [
+              npins
+            ];
+            text = ''
+              echo "<details><summary>npins changes</summary>"
+              # Needed because GitHub's rendering of the first body line breaks down otherwise
+              echo ""
+              echo '```'
+              npins update --directory "$1/npins" 2>&1
+              echo  '```'
+              echo "</details>"
+            '';
           };
-        in
-        ''
-          REPO_ROOT=$1
-          echo "Run automated updates"
-        ''
-        + pkgs.lib.concatStrings (pkgs.lib.mapAttrsToList (title: command: ''
-          echo -e '<details><summary>${title}</summary>\n\n```'
-          ${command} 2>&1
-          echo -e '```\n</details>'
-        '') commands);
-    };
+          cargo = pkgs.writeShellApplication {
+            name = "update-cargo";
+            runtimeInputs = with pkgs; [
+              cargo
+            ];
+            text = ''
+              echo "<details><summary>cargo changes</summary>"
+              # Needed because GitHub's rendering of the first body line breaks down otherwise
+              echo ""
+              echo '```'
+              cargo update --manifest-path "$1/Cargo.toml" 2>&1
+              echo  '```'
+              echo "</details>"
+            '';
+          };
+        };
+      in
+      pkgs.writeShellApplication {
+        name = "auto-pr-update";
+        text = ''
+          # Prevent impurities
+          unset PATH
+          ${lib.concatMapStringsSep "\n" (script: ''
+            echo >&2 "Running ${script}"
+            ${lib.getExe script} "$1"
+          '') (lib.attrValues updateScripts)}
+        '';
+      };
 
     # Tests the tool on the pinned Nixpkgs tree, this is a good sanity check
     nixpkgsCheck = pkgs.runCommand "test-nixpkgs-check-by-name" {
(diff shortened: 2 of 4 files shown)


infinisil:
I renamed the `check` job to `build` because it's more accurate considering what it does. I'll have to update the branch protection to change the checks which are required before we can merge it.

infinisil:
Nice: https://github.com/NixOS/nixpkgs-check-by-name/actions/runs/8636542602/attempts/1#summary-23676696374

willbush (approved):

.github/workflows/main.yml
```diff
@@ -14,3 +14,23 @@ jobs:
...
+      - name: test update script
+        run: |
+          nix-build -A autoPrUpdate
+          {
+            result/bin/auto-pr-update .
+            echo ""
+            echo '```diff'
+            git diff
+            echo '```'
+          } > $GITHUB_STEP_SUMMARY
```
philiptaron, line 34: This looks super nice.

default.nix
```diff
@@ -32,8 +33,12 @@ let
 
 
   results = {
-    build = pkgs.callPackage ./package.nix {
-      inherit nixpkgsLibPath initNix runtimeExprPath testNixpkgsPath;
+    # We're using this value as the root result. By default, derivations expose all of their
+    # internal attributes, which is very messy. We prevent this using lib.lazyDerivation
+    build = lib.lazyDerivation {
```
philiptaron, line 38: TIL about [`lazyDerivation`](https://noogle.dev/f/lib/derivations/lazyDerivation) 👀

infinisil:
Awesome!

Not sure if this needs to be documented anywhere since nobody will actually need to run this locally, but to make it work locally you need a Docker daemon running, which on NixOS means something like this:

```nix
{
  virtualisation.docker.enable = true;
  networking.firewall.trustedInterfaces = [ "docker0" ];
  users.users.me.extraGroups = [ "docker" ];
}
```

scripts/update-github-actions.sh
```diff
@@ -0,0 +1,40 @@
...
+set -euo pipefail
+
+REPO_ROOT=$1
+
+echo -e "<details><summary>GitHub Action updates</summary>\n\n"
+
+# Each dependabot update call tries to update all dependencies,
+# but the resulting files are output individually for each (with the intention of creating a PR for each).
+# We want to have all changes together though, so we just repeatedly take the first one
+# until there's none anymore (-e jq flag)
```
philiptaron, line 16: That's not quite what `-e` does by my reading of [the manual](https://jqlang.github.io/jq/manual/#invoking-jq):

> > `--exit-status / -e:`
>
> Sets the exit status of jq to 0 if the last output value was neither false nor null, 1 if the last output value was either false or null, or 4 if no valid result was ever produced. Normally jq exits with 2 if there was any usage problem or system error, 3 if there was a jq program compile error, or 0 if the jq program ran.

willbush:
We're you using https://github.com/nektos/act ? (it depends on docker)

scripts/update-github-actions.sh
```diff
@@ -0,0 +1,40 @@
...
+while
+  # Unused argument would be the remote GitHub repo, which is not used if we pass --local
+  create_pull_request=$(LOCAL_GITHUB_ACCESS_TOKEN=$(gh auth token) \
+    dependabot update github_actions this-argument-is-unused --local "$REPO_ROOT" \
+    | jq -ecs 'map(select(.type == "create_pull_request")) | .[0].data')
```
philiptaron, line 21: Wow, this is one heck of a pipeline.

I might:

1. Extract `gh auth token` as it won't change in the loop
2. Use the longer form version of `jq` arguments (`--exit-status`, `--compact-output`, `--slurp`)

It's serviceable as is, though.

infinisil:
Nah I'm running it locally by doing what the action does: https://github.com/NixOS/nixpkgs-check-by-name/blob/44d5e2e905037c4614095228d2b9b91eaa119cf2/.github/workflows/update.yml#L19-L20

Or within the current directory:
```
$ nix-build -A autoPrUpdate
$ result/bin/auto-pr-update . > body 
```

scripts/update-github-actions.sh
```diff
@@ -0,0 +1,40 @@
...
+  echo ""
+
+  jq -er '."pr-body"' <<< "$create_pull_request"
+  echo '</details>'
+
+  jq -c '."updated-dependency-files"[]' <<< "$create_pull_request" \
+    | while read -r fileUpdate; do
+      file=$(jq -er '.name' <<< "$fileUpdate")
+      # -j makes sure to not output a trailing newline
+      jq -ejr '.content' <<< "$fileUpdate" > "$REPO_ROOT/$file"
```
philiptaron, line 36: `-j` is `--join-output`

scripts/update-github-actions.sh
```diff
@@ -0,0 +1,40 @@
...
+# We want to have all changes together though, so we just repeatedly take the first one
+# until there's none anymore (-e jq flag)
+while
+  # Unused argument would be the remote GitHub repo, which is not used if we pass --local
+  create_pull_request=$(LOCAL_GITHUB_ACCESS_TOKEN=$(gh auth token) \
+    dependabot update github_actions this-argument-is-unused --local "$REPO_ROOT" \
+    | jq -ecs 'map(select(.type == "create_pull_request")) | .[0].data')
+do
+  title=$(jq -er '."pr-title"' <<< "$create_pull_request")
+  echo "<details><summary>$title</summary>"
```
philiptaron, line 24: In the other scripts, you do: `echo -e '<details><summary>$title</summary>\n\n'` (with two newlines)

philiptaron (approved):
Dang Silvan, I'm learning too much. This is definitely effortful in a way that's quite impressive.

scripts/update-github-actions.sh
infinisil: Good eye, I was kind of lazy with the wording there, will improve that!

scripts/update-github-actions.sh
infinisil: I think it's cleaner here, I'll change the other scripts to match this one

scripts/update-github-actions.sh
infinisil: [(diff)](https://github.com/NixOS/nixpkgs-check-by-name/compare/08e962950f372595027bcbf898a012a97b62d96b..815a4ecd3031e43751baf65c0b2722a6346c6486#diff-a6887f32bd809f67897cc1dd3164202dfe5a6d4933b6f736b3ae864e00a70893R17-R32)

scripts/update-github-actions.sh
infinisil: Done :)

philiptaron (approved):
# 🦺 👍🏻 👀 
Looks good.

philiptaron:
Hrmph. https://github.com/NixOS/nixpkgs-check-by-name/actions/runs/8637321506/job/23679230914

philiptaron:
Lazy derivation didn't work as expected?

infinisil:
Ohh it did work as expected, this is actually great because it shows that we relied on `.version`, which was kind of internal. I'll make a PR to explicitly expose it

infinisil:
Just opened https://github.com/dependabot/cli/issues/301 to hopefully have a better solution eventually :)

infinisil:
Nice: https://github.com/NixOS/nixpkgs-check-by-name/pull/42

PR #41, opened 2024-04-10
Explicitly expose version to avoid release script failure

infinisil:
lazyDerivation limits which attributes are exposed, and version is not currently one of them, so we need to explicitly expose it.

This fixes this CD failure:

https://github.com/NixOS/nixpkgs-check-by-name/actions/runs/8637321506/job/23679230914

This was accidentally introduced in #40 (should've made a separate PR for the first commit after all!)

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -31,13 +31,15 @@ let
     nix-store --init
   '';
 
+  # Determine version from Cargo.toml
+  version = (lib.importTOML ./Cargo.toml).package.version;
 
   results = {
     # We're using this value as the root result. By default, derivations expose all of their
     # internal attributes, which is very messy. We prevent this using lib.lazyDerivation
     build = lib.lazyDerivation {
       derivation = pkgs.callPackage ./package.nix {
-        inherit nixpkgsLibPath initNix runtimeExprPath testNixpkgsPath;
+        inherit nixpkgsLibPath initNix runtimeExprPath testNixpkgsPath version;
       };
     };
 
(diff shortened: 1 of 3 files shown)


philiptaron (approved):
Sure, definitely one way to do it. Note the other set of things being returned at the end of the file. I sort of assumed that's where you'd land the version.

infinisil:
Oh I like that idea more, lemme try something..

infinisil:
Done, I like it more like this :D

default.nix
```diff
@@ -31,15 +31,18 @@ let
     nix-store --init
   '';
 
+  # Determine version from Cargo.toml
+  version = (builtins.fromTOML (builtins.readFile ./Cargo.toml)).package.version;
```
philiptaron, line 35:
```suggestion
  version = (lib.importTOML ./Cargo.toml).package.version;
```

default.nix
```diff
@@ -31,15 +31,18 @@ let
...
+
+  # The build with all internal derivation attributes exposed
+  buildUnredacted = pkgs.callPackage ./package.nix {
+    inherit nixpkgsLibPath initNix runtimeExprPath testNixpkgsPath version;
+  };
```
philiptaron, line 40: I'd just leave this in `lazyDerivation` now that it isn't used elsewhere.

default.nix
```diff
@@ -132,4 +135,6 @@ results.build // results // {
   # Built by CI
   ci = pkgs.linkFarm "ci" results;
 
+  # Used by CI to determine whether a new version should be released
+  inherit version;
```
philiptaron, line 139: Nice.

philiptaron (approved):
More stuff to implement if you want.

infinisil:
Nice suggestions, done!

philiptaron (approved):
# 🚀 

Tight and minimal.

PR #42, opened 2024-04-10
Simplify GitHub action update script

infinisil:
With the advice from https://github.com/dependabot/cli/issues/301#issuecomment-2048440081. Way less hacky this way!

Was originally introduced in #40

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -92,6 +92,7 @@ let
               dependabot-cli
               jq
               github-cli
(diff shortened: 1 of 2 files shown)


philiptaron (approved):
So much nicer.

philiptaron:
> `mktemp: command not found`

Can't have everything, I guess 😆

PR #44, opened 2024-04-15
Fix output markdown of update-github-actions

infinisil:
Previously the closing `</details>` tag was not printed if there was nothing to update, which leads to wrong nesting. See the first revision of the description in https://github.com/NixOS/nixpkgs-check-by-name/pull/43


scripts/update-github-actions.sh
```diff
@@ -38,20 +38,19 @@ job:
...
 
-echo '</details>'
+  jq --compact-output '."updated-dependency-files"[]' <<< "$create_pull_request" \
+    | while read -r fileUpdate; do
+      file=$(jq --exit-status --raw-output '.name' <<< "$fileUpdate")
+      # --join-output makes sure to not output a trailing newline
+      jq --exit-status --raw-output --join-output '.content' <<< "$fileUpdate" > "$REPO_ROOT/$file"
+    done
+fi
+# Nothing to update otherwise
```
philiptaron, line 54:
```suggestion
```

philiptaron (approved):
Comment seems excess to requirements now.

scripts/update-github-actions.sh
infinisil: Done :)

philiptaron (approved):

PR #47, opened 2024-04-17
Format Rust, Nix and Bash via treefmt

infinisil:
Prompted by https://github.com/NixOS/nixpkgs-check-by-name/pull/46#discussion_r1566745927 (cc @willbush). I didn't use a bash formatter, but https://github.com/lovesegfault/beautysh seems to almost match what I've been doing!

For Nix formatting we obviously use https://github.com/NixOS/nixfmt :)

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,12 @@
+# This file contains a list of commits that are not likely what you
+# are looking for in a blame, such as mass reformatting or renaming.
+# You can set this file as a default ignore file for blame by running
+# the following command.
+#
+# $ git config blame.ignoreRevsFile .git-blame-ignore-revs
+
+# Reformatted all Nix files
+81d712a6c4e020bf545fb931b3106aff64871962
+
+# Reformatted all Bash scripts
+b1306dd5ae96b0fb934256d91708733c1f850016
--- a/default.nix
+++ b/default.nix
@@ -4,6 +4,7 @@ in
 {
   system ? builtins.currentSystem,
   nixpkgs ? sources.nixpkgs,
+  treefmt-nix ? sources.treefmt-nix,
 }:
 let
   pkgs = import nixpkgs {
@@ -34,6 +35,13 @@ let
   # Determine version from Cargo.toml
   version = (lib.importTOML ./Cargo.toml).package.version;
 
+  treefmtEval = (import treefmt-nix).evalModule pkgs {
+    # Used to find the project root
+    projectRootFile = ".git/config";
+
+    programs.rustfmt.enable = true;
+  };
+
   results = {
     # We're using this value as the root result. By default, derivations expose all of their
     # internal attributes, which is very messy. We prevent this using lib.lazyDerivation
@@ -51,8 +59,17 @@ let
       nativeBuildInputs = with pkgs; [
         npins
         rust-analyzer
+        treefmtEval.config.build.wrapper
       ];
     };
+
+    # This checks that all Git-tracked files are formatted appropriately
+    treefmt = treefmtEval.config.build.check (
+      lib.fileset.toSource {
+        root = ./.;
+        fileset = lib.fileset.gitTracked ./.;
+      }
+    );
 
     # Run regularly by CI and turned into a PR
     autoPrUpdate =
--- a/default.nix
+++ b/default.nix
@@ -41,6 +41,8 @@ let
 
     programs.rustfmt.enable = true;
     programs.nixfmt-rfc-style.enable = true;
+    programs.shfmt.enable = true;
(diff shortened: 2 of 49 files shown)


willbush:
Can we add a [.git-blame-ignore-revs](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) which helps clean up `git blame` a lot. It's also useful to add to ones git config ([docs](https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt).

willbush:
> https://github.com/lovesegfault/beautysh

Maintenance status of the repo is a little concerning. We might be able to get close to the same format with these options on https://github.com/mvdan/sh

```
Printer options:

  -i,  --indent uint       0 for tabs (default), >0 for number of spaces
  -bn, --binary-next-line  binary ops like && and | may start a line
  -ci, --case-indent       switch cases will be indented
  -sr, --space-redirects   redirect operators will be followed by a space
  -kp, --keep-padding      keep column alignment paddings
  -fn, --func-next-line    function opening braces are placed on a separate line
```

infinisil:
> > https://github.com/lovesegfault/beautysh
> 
> Maintenance status of the repo is a little concerning.

Ohh I looked at your shfmt suggestion, I thought it was https://github.com/patrickvane/shfmt, which when compared to beautysh looked more concerning. Didn't know that https://github.com/mvdan/sh provides shfmt nowadays, switching to that now!

> We might be able to get close to the same format with these options on https://github.com/mvdan/sh

Turns out at least the tabs vs spaces is nicely handled without extra config because it adheres to our [.editorconfig](https://github.com/NixOS/nixpkgs-check-by-name/blob/main/.editorconfig).

scripts/release.sh
```diff
@@ -61,7 +61,7 @@ if ! release=$(gh api \
   echo "Failed to create release: $release"
 fi
 
-releaseId=$(jq .id <<< "$release")
```
willbush, line 64: I recently learned that `  -sr, --space-redirects   redirect operators will be followed by a space` can retain the original style (which I like a little more)

package.nix
```diff
@@ -30,15 +29,13 @@ rustPlatform.buildRustPackage {
...
     nix
-    rustfmt
     clippy
     makeWrapper
   ];
   env.NIX_CHECK_BY_NAME_EXPR_PATH = "${runtimeExprPath}";
   env.NIX_PATH = "test-nixpkgs=${testNixpkgsPath}:test-nixpkgs/lib=${nixpkgsLibPath}";
   preCheck = initNix;
   postCheck = ''
-    cargo fmt --check
```
willbush, line 41: Can we use `treefmt --fail-on-change` here?

scripts/release.sh
infinisil: Sounds good, done!

package.nix
infinisil: This is now handled by https://github.com/NixOS/nixpkgs-check-by-name/blob/ff95d9b072e1726cf2b88a3e9e1314df32df373b/default.nix#L75-L81 (just added the comment above to make this clearer)

willbush (approved):

default.nix
```diff
@@ -51,18 +68,25 @@ let
...
         rust-analyzer
+        treefmtEval.config.build.wrapper
       ];
     };
 
+    # This checks that all Git-tracked files are formatted appropriately
+    treefmt = treefmtEval.config.build.check (
+      lib.fileset.toSource {
+        root = ./.;
+        fileset = lib.fileset.gitTracked ./.;
```
philiptaron, line 79: Look at that dogfooding!

.git-blame-ignore-revs
```diff
@@ -0,0 +1,12 @@
+# This file contains a list of commits that are not likely what you
+# are looking for in a blame, such as mass reformatting or renaming.
+# You can set this file as a default ignore file for blame by running
+# the following command.
+#
+# $ git config blame.ignoreRevsFile .git-blame-ignore-revs
+
+# Reformatted all Nix files
+81d712a6c4e020bf545fb931b3106aff64871962
```
philiptaron, line 9: 81d712a6c4e020bf545fb931b3106aff64871962

.git-blame-ignore-revs
```diff
@@ -0,0 +1,12 @@
...
+
+# Reformatted all Bash scripts
+b1306dd5ae96b0fb934256d91708733c1f850016
```
philiptaron, line 12: b1306dd5ae96b0fb934256d91708733c1f850016

philiptaron (approved):
# ✅ 
`treefmt-nix` is pretty cool.

PR #51, opened 2024-04-18
Fix cargo missing in dev env

infinisil:
<s>No idea how I didn't notice it missing before..

Must've come from somewhere else :shrug:</s>

Turns out `lazyDerivation` (added in https://github.com/NixOS/nixpkgs-check-by-name/pull/40/commits/3f361a461c259b738ad8914749c0ea24ff0ca018) ended up filtering out .nativeBuildInputs and co., which mkShell's inputsFrom relies on to figure out what to use for the environment.. :facepalm:

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -45,26 +45,27 @@ let
     settings.formatter.shfmt.options = [ "--space-redirects" ];
   };
 
+  buildUnfiltered = pkgs.callPackage ./package.nix {
+    inherit
+      nixpkgsLibPath
+      initNix
+      runtimeExprPath
+      testNixpkgsPath
+      version
+      ;
+  };
+
   results = {
     # We're using this value as the root result. By default, derivations expose all of their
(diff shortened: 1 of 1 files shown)


willbush:
Ah strange. I don't have it installed globally either. In my forked copy of this repo I have it so it's coming from somewhere else in the derivation somehow.

willbush (approved):

infinisil:
Figured it out, `lazyDerivation` was the problem, see the commit message :facepalm:

philiptaron (approved):
I don't think `lazyDerivation` is doing too much for us. The promise of being less messy now seems mostly untrue. This is more code and less easy to understand than the simple exposure of the original derivation.

infinisil:
Yeah agreed 😅

Maybe let's just not bother trying to make `nix-build` work, we can still use `nix-build -A build`

I'll update it shortly

philiptaron (approved):
Yeah, let's do this.

PR #55, opened 2024-04-22
Fix link in CONTRIBUTING.md

willbush:
I didn't notice xref was failing for a valid reason when merging https://github.com/NixOS/nixpkgs-check-by-name/pull/54

Decided to point at the release.sh which publishes the release.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,5 +65,5 @@ Integration tests are declared in [`./tests`](./tests) as subdirectories imitati
 
 Pinned dependencies are [regularly updated automatically](./.github/workflows/update.yml).
 
-Releases are [automatically created](./.github/workflows/release.yml) when the `version` field in [`Cargo.toml`](./Cargo.toml)
+Releases are [automatically created](./.github/workflows/after-release.yml) when the `version` field in [`Cargo.toml`](./Cargo.toml)
 is updated from a push to the main branch.


CONTRIBUTING.md
```diff
@@ -65,5 +65,5 @@ Integration tests are declared in [`./tests`](./tests) as subdirectories imitati
 
 Pinned dependencies are [regularly updated automatically](./.github/workflows/update.yml).
 
-Releases are [automatically created](./.github/workflows/release.yml) when the `version` field in [`Cargo.toml`](./Cargo.toml)
+Releases are [automatically created](./.github/scripts/release.sh) when the `version` field in [`Cargo.toml`](./Cargo.toml)
```
infinisil, line 68:
```suggestion
Releases are [automatically created](./.github/workflows/after-release.yml) when the `version` field in [`Cargo.toml`](./Cargo.toml)
```

infinisil (changes requested):
Ah good catch! I now updated the [branch protection ruleset](https://github.com/NixOS/nixpkgs-check-by-name/rules/541493?ref=refs%2Fheads%2Fmain) to enforce all checks to succeed before PRs can be merged (should've done this after adding the xrefcheck workflow).

The link still isn't right though, I made a suggestion :)

CONTRIBUTING.md
infinisil: Oh I see that you intentionally changed it to `scripts/release.sh`. I think pointing to the workflow is better, since that's what causes the script to run automatically (the script by itself could also be run manually).

willbush: I thought `main.yml` caused release.sh to run?

infinisil (approved):
Force pushed the suggestion.

Thought: It's unfortunate that we can't really use GitHub's web interface to create fixup commits, since we enforce merge commits, so we can't squash those. Maybe we should switch to not requiring merge commits? I do otherwise still like the idea of using merge commits, but I'm willing to be convinced otherwise!

CONTRIBUTING.md
infinisil: Oh damn you're right! I take it back, `release.sh` seems much better to point to now :laughing:

willbush: Lol I clicked auto-merge and it merged it before we could fix it back.

infinisil: Opened https://github.com/NixOS/nixpkgs-check-by-name/pull/57 to fix this

philiptaron:
> Maybe we should switch to not requiring merge commits?

I'm still on team "merge commit." Fixups and force-pushing are a fine workflow, or we could be less bothered by one more commit in the merge if created by the WebUI.

PR #57, opened 2024-04-22
Fix the link in CONTRIBUTING more

infinisil:
It's the main.yml that runs the release.sh script. My mistake in #55, thanks for pointing it out @willbush!


infinisil:
I decided to add links to both the workflow and the script, because why not

willbush (approved):

philiptaron (approved):
Clicked the links, looks good.

PR #59, opened 2024-04-29
Replace from_utf8 with more lenient from_utf8_lossy

willbush:
Noticed we're not using [lossy version](https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8_lossy) of `from_utf8` in one place, but we are in two other places:

```
src/eval.rs
178                  String::from_utf8_lossy(&result.stdout)

src/main.rs
276          let actual_errors = String::from_utf8_lossy(&writer).replace(&expr_path, "src/eval.nix");
```


infinisil (approved):
Nice!

PR #65, opened 2024-05-14
flake.nix: init

philiptaron:
I like flakes. 🤷🏻‍♂️ 

* Added `flake.nix`
  * Wired up the package to the default build output for `nix build`
  * Wired up the shell to the default shell for `nix shell`
  * Wired up the two checks (for formatting and nixpkgs) to `nix flake check`.
  * Added the hydra-built systems as the default systems
* Added `nix flake update` to the update scripts
* Reshaped `default.nix` to accept `pkgs` as an input
* Added `treefmtWrapper` (per the docs on [treefmt-nix](https://github.com/numtide/treefmt-nix?tab=readme-ov-file#flakes)) that fits into `nix fmt`.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -88,6 +88,23 @@ let
     autoPrUpdate =
       let
         updateScripts = {
+          flake = pkgs.writeShellApplication {
+            name = "update-flake";
+            runtimeInputs = with pkgs; [
+              git
+              nix
+            ];
+            text = ''
+              echo "<details><summary>flake.nix changes</summary>"
+              # Needed because GitHub's rendering of the first body line breaks down otherwise
+              echo ""
+              echo '```'
+              cd "$1"
+              nix flake update 2>&1
+              echo  '```'
+              echo "</details>"
+            '';
+          };
           npins = pkgs.writeShellApplication {
             name = "update-npins";
             runtimeInputs = with pkgs; [ npins ];
@@ -96,7 +113,7 @@ let
               # Needed because GitHub's rendering of the first body line breaks down otherwise
               echo ""
               echo '```'
-              npins update --directory "$1/npins" 2>&1
+              npins --directory "$1/npins" import-flake 2>&1
               echo  '```'
               echo "</details>"
             '';
--- a/default.nix
+++ b/default.nix
@@ -5,13 +5,15 @@ in
   system ? builtins.currentSystem,
   nixpkgs ? sources.nixpkgs,
   treefmt-nix ? sources.treefmt-nix,
+  pkgs ? (
+    import nixpkgs {
+      inherit system;
+      config = { };
+      overlays = [ ];
+    }
+  ),
 }:
 let
-  pkgs = import nixpkgs {
-    inherit system;
-    config = { };
-    overlays = [ ];
-  };
   inherit (pkgs) lib;
 
   runtimeExprPath = ./src/eval.nix;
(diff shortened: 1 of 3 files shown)


philiptaron:
Turns out that `nix flake update` is failing in the CI scripts. I wonder if it's a "nix from inside nix" problem. There's no output, which is sad.

infinisil (changes requested):
I'm not against having a flake.nix in principle, but this currently duplicates all the input locking! And I'd rather not use flake inputs because then they can't be overridden by stable Nix (unless you [implement it manually](https://github.com/edolstra/flake-compat/pull/49)).

So `flake.nix` should just be a tiny layer above `default.nix`, like [the one in nixfmt](https://github.com/NixOS/nixfmt/blob/master/flake.nix).

philiptaron:
@infinisil, it _is_ a tiny layer above `default.nix`! I did it differently than the nixfmt one because the Nix code is shaped differently, but there's no getting around the flake's use of flake.lock. I thought the simplest way to go was duplication rather than trying to make `npins` flake compatible.

infinisil:
> there's no getting around the flake's use of flake.lock

There is though! If you just don't specify any flake inputs and instead import the `default.nix` (which then uses the niv-pinned ones) with its default arguments, it should work just fine

philiptaron:
> if you don't specify flake inputs

But that's what I want the flake _for_ -- specifying inputs. Without any inputs, as a flake consumer, I have to use whatever's pinned, and there's no ability to override.

I think the right way to go is to update npins to 0.2.4 (per the linked review), then use the [import-flake](https://github.com/andir/npins/pull/70) to base the npins off flake.lock.

willbush:
Would `.envrc` be `use flake` or `use nix`?

philiptaron:
> Would `.envrc` be `use flake` or `use nix`?

It would be able to contain `use flake` should this PR be merged.

infinisil:
> Without any inputs, as a flake consumer, I have to use whatever's pinned, and there's no ability to override.

You can override the pinned sources using the `default.nix`. E.g.

```nix
{
  inputs.ncbn.url = "github:NixOS/nixpkgs-check-by-name";
  inputs.ncbn.flake = false;

  outputs = { ncbn, ... }:
    let
      result = import ncbn {
        nixpkgs = <something else>;
      };
    in
    {
      packages.ncbn = result.build;
    };
}
```

This notably has the advantage that you can even override it with a version of Nixpkgs that Flake inputs can't represent, e.g. one patched via a derivation, or fetched with a non-builtin fetcher, etc.

willbush:
Feel like I don't know enough about the pros/cons of flakes despite using them for like 4 years.

I think at a minimum we shouldn't have two lock files for the same inputs.

Is there any tradeoff in evaluation caching between using flakes vs npins lock file?

philiptaron:
I'm returning to this PR after a couple weeks, but I'm going to be experimenting with how GitHub actions runs the code a bit so please bear it no mind until it's out of draft.

willbush:
> > Would `.envrc` be `use flake` or `use nix`?
> 
> It would be able to contain `use flake` should this PR be merged.

I noticed [this](https://github.com/helix-editor/helix/blob/6d64e6288add793caf1c841ea295518414c04ea0/.envrc#L6) recently `use flake || use nix`.

PR #79, opened 2024-07-08
Allow nixpkgs-check-by-name to work with recent Nix versions

### LINKED ISSUE #78: nixpkgs-check-by-name reports IO error for Nix versions 2.21+
When running `nixpkgs-check-by-name` with Nix versions greater than 2.20, it reports the following:

```
I/O error:  Failed to get the definition info for attribute acl: In pkgs/stdenv/linux/default.nix, attribute parent node is not an attribute path node: NODE_INHERIT@28432..28603
```

philiptaron:
Fixes #78. I didn't root cause what the new behavior in the most recent Nix versions was that made the output of nix-instantiate different.


infinisil:
Nice, this looks like the right fix! Could you also extend the test suite I mentioned in https://github.com/NixOS/nixpkgs-check-by-name/issues/78#issuecomment-2224437515 with a case for this? Furthermore it would be great to have some comments giving the extra context

test-nixpkgs-check-by-name.nix
infinisil, line 1: Nit: Let's avoid the redundancy and just call it `test-on-nixpkgs.nix` or `nixpkgs-check.nix` or so :)

test-nixpkgs-check-by-name.nix
```diff
@@ -0,0 +1,38 @@
...
+    runCommand "test-nixpkgs-check-by-name-with-${nix.name}"
+      {
+        nativeBuildInputs = [
+          nixpkgs-check-by-name
+          nix
+        ];
+        nixpkgsPath = nixpkgs;
+
+        # Allow running against all other Nix versions.
+        passthru = lib.mapAttrs (name: value: self value) nixDerivations;
```
infinisil, line 30:
```suggestion
        # Allow running against all other Nix versions.
        passthru.nixVersions = lib.mapAttrs (name: value: self value) nixVersions;
```

`passthru` attributes should always be static, so that there's no namespace overlap with other resulting derivation attributes.

And the `tryEval` mapping above is not necessary :)

default.nix
```diff
@@ -157,21 +157,15 @@ let
...
-        '';
+    # Tests the tool on the pinned Nixpkgs tree with stable Nix. This is a good sanity check.
+    # There's a `passthru` attribute that allows testing this on every other Nix version in nixpkgs.
+    nixpkgsCheck = pkgs.callPackage ./test-nixpkgs-check-by-name.nix {
+      inherit initNix nixpkgs;
+      nixpkgs-check-by-name = packages.build;
+    };
+
+    # Tests the tool on the pinned Nixpkgs tree with the most recent Nix. This is a sanity check.
+    nixpkgsCheckWithLatestNix = packages.nixpkgsCheck.passthru.latest;
```
infinisil, line 168:
```suggestion
    nixpkgsCheckWithLatestNix = packages.nixpkgsCheck.nixVersions.latest;
```

`passthru`'s get "passed through" as a derivation attribute, so there's no need to refer to it via `passthru`

infinisil (changes requested):

test-nixpkgs-check-by-name.nix
philiptaron: > And the tryEval mapping above is not necessary :)

It is if you want to run against everything in the nixVersions array. Otherwise:

```
       error: nix_2_10 has been removed
```

default.nix
philiptaron: The use of the inner `nixVersions` attrset resolved the reason I was using `passthru` directly here.

test-nixpkgs-check-by-name.nix
philiptaron: Sure. I named it based on the existing package name, but I'll follow whatever.

philiptaron:
> Could you also extend the test suite I mentioned in https://github.com/NixOS/nixpkgs-check-by-name/issues/78#issuecomment-2224437515 with a case for this?

I'm trying, but I have no idea how you generated that test case. 😬 Could you help me out with reproduction instructions?

test-nixpkgs-check-by-name.nix
infinisil: I don't think we should worry about running it for all Nix version attributes, instead limiting it to `nixVersions.{minimum,stable,latest,git}` should be good enough.

default.nix
infinisil: (not sure what you mean here or why it's resolved)

infinisil:
Oh yeah I can't quite remember how I got those numbers, probably just some trial-and-error, but obviously without actually running the final version through Nix eval, because the file in main is invalid :P

Your approach is pretty good though! I'm suggesting some changes:

```nix
testG: {
  testA = 1;
  "testB" = 2;
  ${"testC"} = 3;
  "${"testD"}" = 4;

  # A
  testE
  # B
  =
  # C
  5
  # D
  ;
  # E

  /**/testF/**/=/**/5/**/;/*E*/

  inherit testG;
  inherit ({}) testH;

  inherit ({})
    testI testJ
    testK testL;
}
```

```
nix run nixpkgs#nixVersions.latest -- eval --expr 'let value = import ./test.nix null; in builtins.mapAttrs (n: _: { inherit (builtins.unsafeGetAttrPos n value) line column; }) value' --impure --json | jq -r .
```

test-nixpkgs-check-by-name.nix
philiptaron: I've kept the mapping, as it's quite useful to me to be able to run against all Nix and Lix versions, but I have added a top-level value -- which runs in CI -- for the following:

1. Nix minimum
2. Nix latest
3. Nix git
4. Lix stable
5. Lix latest

philiptaron:
I'll do the requested version bump (and hence release) in a separate PR.

src/nix_file.rs
```diff
@@ -471,42 +483,77 @@ mod tests {
...
+            (4, 3, Right("${\"testC\"} = 3;")),
+            (5, 3, Right("\"${\"testD\"}\" = 4;")),
+            (8, 3, Right("testE\n  # B\n  =\n  # C\n  5\n  # D\n  ;")),
+            (17, 7, Right("testF/**/=/**/5/**/;")),
+            (19, 11, Left("testG")),
+            (20, 16, Left("testH")),
+            (23, 5, Left("testI")),
+            (23, 11, Left("testJ")),
+            (24, 5, Left("testK")),
+            (24, 11, Left("testL")),
```
infinisil, line 536: These cases need to be run for the locations reported by various Nix versions, such that it works for all of them and not just a single one. Something like

src/nix_file.rs
```diff
@@ -190,9 +190,7 @@ impl NixFile {
             // Something like `foo`, `"foo"` or `${"foo"}`
```
infinisil, line 190: While testing, I found out that this is the branch also used for the `foo` in `inherit (bar) foo` in Nix > 2.20.

```suggestion
            // Something like `foo`, `"foo"` or `${"foo"}`
            // For Nix versions > 2.20, this is also `foo` in `inherit (bar) foo`
```

src/nix_file.rs
```diff
@@ -190,9 +190,7 @@ impl NixFile {
...
         } else if ast::Inherit::can_cast(node.kind()) {
             // Something like `inherit <attr>` or `inherit (<source>) <attr>`.
```
infinisil, line 192:
```suggestion
            // Only for Nix versions <= 2.20: Something like `inherit <attr>` or `inherit (<source>) <attr>`.
```

src/nix_file.rs
```diff
@@ -210,6 +208,12 @@ impl NixFile {
             )
         };
 
+        if ast::Inherit::can_cast(attrpath_node.kind()) {
+            // Something like `inherit <attr>` or `inherit (<source>) <attr>`.
+            // We only look for ones like `<attr-path> = <value>`, so ignore this.
+            return Ok(Left(node.to_string()));
+        };
+
```
infinisil, line 216:
```suggestion
        // This can only happen for Nix versions > 2.20, see the above `Attr` cast
        if ast::Inherit::can_cast(attrpath_node.kind()) {
            // Something like `inherit <attr>` or `inherit (<source>) <attr>`.
            // We only look for ones like `<attr-path> = <value>`, so ignore this.
            return Ok(Left(node.to_string()));
        };

```

infinisil (changes requested):
Out of time for now, but ended up spending a bunch of time to improve the testing with #83

default.nix
```diff
@@ -158,21 +158,20 @@ let
...
+      nix = pkgs.nixVersions.stable;
+    };
+
+    # Tests the tool on the pinned Nixpkgs tree with various Nix and Lix versions.
+    # This allows exposure to changes in behavior from Nix and Nix-alikes.
+    nixpkgsCheckWithLatestNix = packages.nixpkgsCheck.nixVersions.latest;
+    nixpkgsCheckWithGitNix = packages.nixpkgsCheck.nixVersions.git;
+    nixpkgsCheckWithMinimumNix = packages.nixpkgsCheck.nixVersions.minimum;
+    nixpkgsCheckWithStableLix = packages.nixpkgsCheck.lixVersions.stable;
+    nixpkgsCheckWithLatestLix = packages.nixpkgsCheck.lixVersions.latest;
```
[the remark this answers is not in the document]
philiptaron: This is way less important now that #83 is merged and the tool doesn't depend on the ambient version of Nix. I'm fine cutting them all, or going down to {minimum,stable,latest}.

That said, given that reviews are usually a matter of a week or more, 3.5 minutes more isn't an actual issue. Amdahl's law!

src/nix_file.rs
philiptaron: With #83 this is now possible. I'll update.

philiptaron:
I'll take the various suggestions about comments and fix it up to run the cargo tests under a plethora of Nix versions.

philiptaron:
Nix eval (as performed by `nixpkgs-check-by-name`) is currently broken in `nixpkgs` master. https://github.com/NixOS/nixpkgs/pull/329212 fixes it.

philiptaron:
This is drafted until an update for nixpkgs causes eval on 2.3.18 to pass.

default.nix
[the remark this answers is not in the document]
philiptaron: I decided to keep these in the CI run. The number of PRs that aren't you or I is quite low.

infinisil (approved):
Let's go! Thanks for this :D

PR #83, opened 2024-07-19
Fixed Nix version and testing of multiple versions (and fix them for nixVersions.minimum)

infinisil:
This is a precursor to #79:
- Use a fixed Nix version instead of whatever happens to be in `PATH`, so we shouldn't ever have issues like #78 anymore
- Sets up testing for other Nix versions as part of the main build. In comparison with #79, this runs the entire test suite with multiple Nix versions, giving both more coverage while also taking much less time.
- Fixes the build for `nixVersions.minimum`, which we should switch to in the future, because it's what Nixpkgs CI should be using.

This also makes it very easy to add further Nix versions to be tested, which is what #79 can then benefit from.

---

This work is sponsored by [Antithesis](https://antithesis.com/) :sparkles:

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -45,6 +45,9 @@ let
     settings.formatter.shfmt.options = [ "--space-redirects" ];
   };
 
+  # The resulting package is built to always use this Nix version, such that the result is reproducible
+  defaultNixPackage = pkgs.nix;
+
   packages = {
     build = pkgs.callPackage ./package.nix {
       inherit
@@ -54,10 +57,12 @@ let
         testNixpkgsPath
         version
         ;
+      nix = defaultNixPackage;
     };
 
     shell = pkgs.mkShell {
       env.NIX_CHECK_BY_NAME_EXPR_PATH = toString runtimeExprPath;
+      env.NIX_CHECK_BY_NAME_NIX_PACKAGE = lib.getBin defaultNixPackage;
       env.NIX_PATH = "test-nixpkgs=${toString testNixpkgsPath}:test-nixpkgs/lib=${toString nixpkgsLibPath}";
       env.RUST_SRC_PATH = "${pkgs.rustPlatform.rustLibSrc}";
       inputsFrom = [ packages.build ];
@@ -69,6 +74,7 @@ let
         rust-analyzer
         rustfmt
         treefmtEval.config.build.wrapper
+        defaultNixPackage
       ];
     };
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -181,6 +181,7 @@ mod tests {
     use crate::process;
     use crate::utils;
     use anyhow::Context;
+    use std::ffi::OsStr;
     use std::fs;
     use std::path::Path;
     use tempfile::{tempdir_in, TempDir};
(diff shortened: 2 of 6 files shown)


default.nix
```diff
@@ -45,6 +45,10 @@ let
     settings.formatter.shfmt.options = [ "--space-redirects" ];
   };
 
+  # The resulting package is built to always use this Nix version, such that the result is reproducible
+  # TODO: Switch this to pkgs.nixVersions.minimum, because that's what Nixpkgs CI should use
```
philiptaron, line 49: I'm interested in the reasoning behind this "should". My experience is that it's better if CI uses the most recent version of its dependencies, since that usually means that fixes can occur at the right level. Using the minimum version, in my experience, leads to hard-to-remove workarounds for old bugs, which don't contribute to a better product. And the process of upgrading the minimum version has a nasty habit of finding odd new local minima, which then need to be worked around again.

The current Nixpkgs way of doing things -- having a stable version, which moves forward on a ~six monthly cadence in normal times -- is my preferred CI set up.

infinisil:
This does increase the closure size (gzip-compressed from 16MB to 33MB), because a Nix version is now included, but I value reproducibility more, so I think that's fine :)

src/main.rs
```diff
@@ -277,14 +278,24 @@ mod tests {
         } else {
             Path::new("tests/empty-base")
         };
+        // Empty dir, needed so that no warnings are printed when testing older Nix versions
+        // that don't recognise certain newer keys in nix.conf
+        let nix_conf_dir = tempdir()?;
```
philiptaron, line 283: A cute way to do this would be to take a dependency on `pkgs.emptyDirectory`.

philiptaron (approved):
Approved with a disagreement about the CI version TODO.

philiptaron:
> This does increase the closure size (gzip-compressed from 16MB to 33MB), because a Nix version is now included, but I value reproducibility more, so I think that's fine :)

Yeah, it's in the noise.

default.nix
infinisil: Oh yeah, we totally shouldn't have to use the minimum version here. The minimum version should only be used for the part of CI that checks that everything evaluates, which is ofborg (though that [actually use the minimum version](https://github.com/NixOS/ofborg/issues/612)!).

Removed the TODO now :)

src/main.rs
infinisil: Agreed, but since we're not in Nix here, it would be a bit cumbersome, so I think it's fine as-is :P

infinisil:
Ohh, we should be able avoid having to [install Nix](https://github.com/NixOS/nixpkgs/blob/7253197f87ceca94f0f9ce009efe4b153a9e0bfc/.github/workflows/check-by-name.yml#L97-L98) in Nixpkgs GitHub workflow then!

src/main.rs
philiptaron: Totally.

philiptaron:
> Ohh, we should be able avoid having to [install Nix](https://github.com/NixOS/nixpkgs/blob/7253197f87ceca94f0f9ce009efe4b153a9e0bfc/.github/workflows/check-by-name.yml#L97-L98) in Nixpkgs GitHub workflow then!

No, I think you still need it for the `nix-store --import` line below that, I think.

philiptaron (approved):

PR #84, opened 2024-07-20
eval: use a work directory instead of an environment variable to control the Nix expression

philiptaron:
# Motivation

This reduces the list of environment variables needed in order to use the unwrapped `nixpkgs-check-name` tool to one: `NIX_CHECK_BY_NAME_NIX_PACKAGE`. #79 now needs to use the unwrapped tool in order to control which Nix is used in an integration test, but it shouldn't have to figure this detail out.

No test used the `NIX_CHECK_BY_NAME_EXPR_PATH` to change the Nix file that was read, and it's simpler and shorter to rebuild in order to test out changes to the `src/eval.nix`.

# Things Done

- [x] Removed the Nix plumbing around `runtimeExprPath`
- [x] Removed the `NIX_CHECK_BY_NAME_EXPR_PATH` environment variable
- [x] Embedded `eval.nix` in the binary
- [x] Used a work directory instead of a temporary file

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -14,7 +14,6 @@ let
   };
   inherit (pkgs) lib;
 
-  runtimeExprPath = ./src/eval.nix;
   testNixpkgsPath = ./tests/mock-nixpkgs.nix;
   nixpkgsLibPath = nixpkgs + "/lib";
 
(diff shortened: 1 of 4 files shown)


willbush (approved):
LGTM, but I'd wait for @infinisil review. He has a much better understanding of this than I do.

infinisil (approved):
Nice, LGTM!

PR #86, opened 2024-07-22
eval: remove is_test and embed mock-nixpkgs.nix

philiptaron:
# Motivation

Similar to #84, reducing the amount of needed machinery in Nix in favor of Rust code. In particular, remove `NIX_PATH` manipulation in Nix derivations!

# Things Done

This commit does the following things:

1. Purify the environment variables entirely when calling `nix-instantiate`. Opt in to the five environment variables needed in order to run `nix-instantiate` in the Nix sandbox.
2. Remove the hard-coded `NIX_PATH` environment variables in `default.nix` and `package.nix`. These were implicitly sent through to `nix-instantiate`, along with `-I` arguments. Now, only `-I` is used.
3. Replace the `is_test` argument with `#[cfg(test)]` applied to `mutate_nix_instatiate_arguments_based_on_cfg`.

While this is slightly more lines of code that the combination of Nix and Rust that it replaces, there is less Nix logic and more Rust logic.

Additionally, the Rust logic that is test-only is contained to `cargo check` only rather than being present in the production binary.

### DIFF
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -11,13 +11,11 @@ use crate::validation::ResultIteratorExt as _;
 use crate::validation::{self, Validation::Success};
 use crate::NixFileStore;
 use relative_path::RelativePathBuf;
-use std::fs;
-use std::path::Path;
+use std::path::{Path, PathBuf};
+use std::{env, fs, process};
 
 use anyhow::Context;
 use serde::Deserialize;
-use std::path::PathBuf;
-use std::process;
 use tempfile::Builder;
 
 const EVAL_NIX: &[u8] = include_bytes!("eval.nix");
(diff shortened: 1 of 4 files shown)


src/eval.rs
```diff
@@ -100,14 +98,63 @@ pub enum DefinitionVariant {
...
 
+/// Pass through variables needed to make Nix evaluation work inside Nix build. See `initNix`.
+/// If these variables don't exist, assume we're not in a Nix sandbox.
+fn pass_through_environment_variables_for_nix_eval_in_nix_build(command: &mut process::Command) {
+    for variable in [
+        "NIX_CONF_DIR",
+        "NIX_LOCALSTATE_DIR",
+        "NIX_LOG_DIR",
+        "NIX_STATE_DIR",
+        "NIX_STORE_DIR",
```
infinisil, line 109: Note for the future: These probably aren't all needed, because we don't do any building, just evaluation. I suspect only `NIX_STATE_DIR` is necessary

infinisil (approved):
Very neat, this is much cleaner!

PR #46, opened 2024-04-16
Automated scheduled releases with changelogs [WIP]

infinisil:
This adds a fully automated changelog and version pipeline, fixing #38. I looked into mainly Towncrier and Changie, but both of them weren't super integrated, so I wrote my own solution, which ended up being fairly simple.

The solution here works like this:
- Every PR has to have a "This change is user-facing" checkbox to the PR description (checked in the default PR template), and either:
  - Uncheck it to indicate it's not a user-facing change
  - Or add a changelog by creating a file in one of `changes/unreleased/{major,medium,minor}`.
  
    The file has to be a markdown file with its first line like `#` (and no other lines starting with `#`).

- Every week on Tuesday (presumably after the automated update PR from Monday is merged, but can also be triggered manually at any point), a version PR is opened if the main branch has any unreleased changelogs. This PR includes:
  - Depending on whether there are any major, medium or only minor changelogs, the appropriate version number is bumped in `Cargo.toml`
  - A new `changes/released/$MAJOR.$MEDIUM.$MINOR.md` file that compiles all previously unreleased changelogs into a single file. Each changelog entry is a Markdown list entry of the form
    ```markdown
    - <title> by @<author> in #<PR>
      <optional rest of lines>
    ```
    
    The author and PR number is automatically determined based on which PR added the changelog file.

- Once that PR is merged (optionally with manual updates), the release will be published, with the compiled changelog included.

  Currently there's nothing that would ever update the changelog of releases, but I don't expect that to be necessary.

This PR needs more testing and comments before it can be merged.

- [ ] This change is user-facing

---

This work is sponsored by [Antithesis](https://antithesis.com/) :sparkles:

PR #89, opened 2024-07-26
Version 0.1.2

infinisil:
Haven't had a release in a while, but we need one now because we have (multiple!) fixes for https://github.com/NixOS/nixpkgs-check-by-name/issues/78 :D

The release will be created automatically when this is merged, but we don't have a way for nice changelog generation (still want to finish #46! Should be almost done really), but the main thing to know is just that #78 is fixed, which allows us to un-revert https://github.com/NixOS/nixpkgs/pull/329436 in Nixpkgs

Thanks for the contributions for this release, @philiptaron and @willbush!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "nixpkgs-check-by-name"
-version = "0.1.1"
+version = "0.1.2"
 edition = "2021"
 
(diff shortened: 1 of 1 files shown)


philiptaron (approved):
# 💪🏻 

Every release is a win.

PR #91, opened 2024-07-27
main: extract a Status object to represent the final status of `nixpkgs-check-by-name`

philiptaron:
# Motivation for changes

I didn't like the mixing of error reporting (through `io::Writer`) and business logic inside the `process` function. By extracting a Status object, the two are separated, just like `NixpkgsProblem` did on a lower level.

# Description of changes

- [x] Extracted `Status` object which contains the actual status of `nixpkgs-check-by-name` checks.
- [x] Isolated coloring decisions to private `Status::fmt` method.
- [x] Isolated ExitCode transformation to `From` impl.
- [x] Introduced `ColoredStatus` which is the colored version of status; it's used exclusively in `main`.
- [x] All tests continue to pass without the `NO_COLOR` environment variable and test-only `io::Writer`.

It's mildly longer (~53 more lines) due to the needed boilerplate of being a module.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,24 +1,24 @@
-use crate::nix_file::NixFileStore;
-use std::panic;
 mod eval;
 mod nix_file;
 mod nixpkgs_problem;
 mod ratchet;
 mod references;
+mod status;
 mod structure;
 mod utils;
 mod validation;
 
+use anyhow::Context as _;
+use clap::Parser;
+use std::path::{Path, PathBuf};
+use std::process::ExitCode;
+use std::{panic, thread};
+
+use crate::nix_file::NixFileStore;
+use crate::status::{ColoredStatus, Status};
 use crate::structure::check_structure;
 use crate::validation::Validation::Failure;
 use crate::validation::Validation::Success;
-use anyhow::Context;
-use clap::Parser;
-use colored::Colorize;
-use std::io;
-use std::path::{Path, PathBuf};
-use std::process::ExitCode;
-use std::thread;
 
 /// Program to check the validity of pkgs/by-name
 ///
@@ -47,14 +47,9 @@ pub struct Args {
 
 fn main() -> ExitCode {
     let args = Args::parse();
-    match process(args.base, args.nixpkgs, &mut io::stderr()) {
-        Ok(true) => ExitCode::SUCCESS,
-        Ok(false) => ExitCode::from(1),
-        Err(e) => {
-            eprintln!("{} {:#}", "I/O error: ".yellow(), e);
-            ExitCode::from(2)
-        }
-    }
+    let status: ColoredStatus = process(args.base, args.nixpkgs).into();
+    eprintln!("{status}");
+    status.into()
 }
 
 /// Does the actual work. This is the abstraction used both by `main` and the tests.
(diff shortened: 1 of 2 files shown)


willbush:
Damn man your killin it today!

src/status.rs
```diff
@@ -0,0 +1,126 @@
...
+    Error(anyhow::Error),
+}
+
+impl Status {
+    fn fmt(&self, f: &mut fmt::Formatter, use_color: bool) -> fmt::Result {
+        let maybe_green = |s: &str| if use_color { s.green() } else { s.normal() };
+        let maybe_yellow = |s: &str| if use_color { s.yellow() } else { s.normal() };
+        let maybe_red = |s: &str| if use_color { s.red() } else { s.normal() };
+
+        let errors = match self {
```
infinisil, line 37:
```suggestion
        let (errors, message) = match self {
```

How about this to avoid the `unreachable!`

src/status.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    /// Some other error occurred.
+    Error(anyhow::Error),
+}
+
+impl Status {
+    fn fmt(&self, f: &mut fmt::Formatter, use_color: bool) -> fmt::Result {
+        let maybe_green = |s: &str| if use_color { s.green() } else { s.normal() };
+        let maybe_yellow = |s: &str| if use_color { s.yellow() } else { s.normal() };
+        let maybe_red = |s: &str| if use_color { s.red() } else { s.normal() };
```
infinisil, line 35: This will cause it to colorise the output even if it's e.g. just a pipe to a file. So I think I prefer the previous code using `NO_COLOR=1` to disable colors when testing, and https://github.com/NixOS/nixpkgs/blob/07c1a652be3a1dda9047f4d8545e17d12958a148/.github/workflows/check-by-name.yml#L113-L115 to force turn it on in CI.

src/main.rs
```diff
@@ -151,39 +93,33 @@ pub fn process<W: io::Write>(
 /// This does not include ratchet checks, see ../README.md#ratchet-checks
 /// Instead a `ratchet::Nixpkgs` value is returned, whose `compare` method allows performing the
 /// ratchet check against another result.
-pub fn check_nixpkgs(nixpkgs_path: &Path) -> validation::Result<ratchet::Nixpkgs> {
+fn check_nixpkgs(nixpkgs_path: &Path) -> validation::Result<ratchet::Nixpkgs> {
+    let nixpkgs_path = nixpkgs_path.canonicalize()?;
```
infinisil, line 97: I think the `with_context` is important to maintain, because it's not obvious which file failed to canonicalise otherwise.

infinisil (changes requested):
This is a nice improvement overall!

src/status.rs
philiptaron: No, it's the same as before.

`nixpkgs-check-by-name` used to unilaterally colorize by calling the `.red`, `.yellow`, etc methods, and in the `colored` crate it would decide to _actually_ not color based on [this environment variable](https://github.com/colored-rs/colored/blob/master/src/control.rs#L98-L101).

Now, it'll only call those methods if `true` is passed to `use_color`. `ColoredStatus` does this -- see the end of the file. `main` uses `ColoredStatus`, and so it retains the original behavior. Tests, which use the `process` method, format a `Status`, which passes `false` to `use_color` in their display, bypassing colorization in this module itself.

src/main.rs
philiptaron: Restored.

src/status.rs
philiptaron: I figured out a way to avoid it using a getter for errors.

infinisil (approved):
Forgot about this, looking good, let's go!

PR #90, opened 2024-07-26
workflows/after-release: Allow manually running

infinisil:
For some reason it didn't trigger for the 0.1.2 release. Let's trigger it manually for now:

image

PR #97, opened 2024-08-24
Better error for outside references

infinisil:
Motivated by @teto in https://github.com/NixOS/nixpkgs-check-by-name/issues/64#issuecomment-2308447186, improves the error messages when a path outside a package directory is referenced:

```
- pkgs/by-name/aa/aa: File package.nix at line 1 contains the path expression "../." which may point outside the directory of that package.
  This is undesirable because it creates dependencies between internal paths, making it harder to reorganise Nixpkgs in the future.
  Alternatives include:
  - If you are creating a new version of a package with a common file between versions, consider following the recommendation in https://github.com/NixOS/nixpkgs/tree/master/pkgs/by-name#recommendation-for-new-packages-with-multiple-versions.
  - If the path being referenced could be considered a stable interface with multiple uses, consider exposing it via a `pkgs` attribute, then taking it as a attribute argument in package.nix.
  - If the path being referenced is internal and has multiple uses, consider passing the file as an explicit `callPackage` argument in `pkgs/top-level/all-packages.nix`.
  - If the path being referenced is internal and will need to be modified independently of the original, consider copying it into the pkgs/by-name/aa/aa directory.
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -258,7 +258,7 @@ dependencies = [
 
 [[package]]
 name = "nixpkgs-check-by-name"
-version = "0.1.2"
+version = "0.1.3"
 dependencies = [
  "anyhow",
  "clap",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "nixpkgs-check-by-name"
-version = "0.1.2"
+version = "0.1.3"
 edition = "2021"
 
 [dependencies]
--- a/src/main.rs
+++ b/src/main.rs
@@ -176,7 +176,7 @@ mod tests {
         test_nixpkgs(
             "case_sensitive",
             path,
-            "pkgs/by-name/fo: Duplicate case-sensitive package directories \"foO\" and \"foo\".\n\
+            "- pkgs/by-name/fo: Duplicate case-sensitive package directories \"foO\" and \"foo\".\n\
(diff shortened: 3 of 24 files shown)


teto:
The `This is undesirable because it creates dependencies between internal paths, making it harder to reorganise Nixpkgs in the future.` part is nice to have. As for the rest, it depends how you prefer to manage the project but one solution is to do like shellcheck and reference a wiki page like "https://www.shellcheck.net/wiki/SC2155" which can be edited without needing another release.

infinisil:
Yeah I'd really like that better, maybe let's keep it simpler for now though

philiptaron (approved):
I kind of hate the `-` prefix, but I suppose consistency is better.

infinisil:
@philiptaron I think it's kind of important for multiple failures: https://github.com/NixOS/nixpkgs-check-by-name/blob/better-error-outside-reference/tests/multiple-failures/expected

infinisil:
The [after release](https://github.com/NixOS/nixpkgs-vet/blob/main/.github/workflows/after-release.yml) action didn't fire automatically yet again, just like [before](https://github.com/NixOS/nixpkgs-vet/pull/90)

infinisil:
I think this is the culprit: https://github.com/orgs/community/discussions/54574#discussioncomment-10119733

Issue #100, opened 2024-08-26
Rename to `nixpkgs-vet`


philiptaron:
The purpose of this tool is to [vet](https://en.wiktionary.org/wiki/vet#Verb) nixpkgs.

There are now, as of late August 2024, 28 specific errors that this tool checks for. Some of them have to do with [RFC 140](https://github.com/NixOS/rfcs/blob/master/rfcs/0140-simple-package-paths.md) and the `pkgs/by-name` directory, but many of them are good hygiene for packages in general that _happen_ to be checked as part of this migration.

As the migration completes, this tool will continue to have a place as a leveraged way to vet the contents of packages that are contributed to Nixpkgs.

We should therefore rename it to the short, punchy name of `nixpkgs-vet`.

willbush:
Oh think I misunderstood what you meant by https://github.com/NixOS/nixpkgs-check-by-name/issues/99#issuecomment-2311162464. Want me to make a PR here to rename?

philiptaron:
Yes please! Then:

1. An org owner can rename this repository
2. We can make a release
3. We can update nixpkgs

philiptaron:
All right, once we do a release, update nixpkgs, and rename the team in the org, we'll be done with this.

infinisil:
Team is renamed: https://github.com/orgs/NixOS/teams/nixpkgs-vet :)

infinisil:
Release can be done by just updating `Cargo.{toml,lock}`

philiptaron:
With #106, I'm calling this done.

infinisil:
Oh nice catch, changed it to "Tool to vet (check) Nixpkgs, including its pkgs/by-name directory" now

PR #104, opened 2024-08-31
Implement package description checking

infinisil:
Kind of an experiment, implements almost all of https://github.com/NixOS/nixpkgs/blob/0abfc619bcb605299a0f3f01c1887bb65db61a6b/pkgs/README.md#L373-L381

#100 is a prerequisite, because this check has nothing to do with `by-name`

Ideally nixpkgs-vet had a configuration file, so we wouldn't have to hardcode these exact checks


philiptaron:
This is pretty great Silvan. Let me know when you're complete in the work enough for a review.

infinisil:
Mainly just drafted this because we depend on the nixpkgs-vet rename to be complete :)

infinisil:
Asked about thoughts on this [on Matrix](https://matrix.to/#/!kjdutkOsheZdjqYmqp:nixos.org/$T9_RAPw7bNA5t1nORrJWiaH-DHCoe3j4NfSuEe5hGVM?via=lossy.network&via=matrix.org&via=nixos.dev), @adisbladis had a good one in that there's already meta checks in Nixpkgs, this completely separate implementation diverges the tooling, makes it confusing.

I'm now thinking that such an external tool should only be used for things that can't be checked with Nix code directly. And @adisbladis mentioned that meta checks apparently aren't very expensive: https://github.com/NixOS/nixpkgs/pull/273935#issuecomment-1854741373

The one thing that's making me hesitant though is the need to implement the [ratched check](https://github.com/NixOS/nixpkgs-vet?tab=readme-ov-file#ratchet-checks) mechanism within Nix, which I'm really not too excited about, and it _is_ working very well with this tool

infinisil:
A neat idea that we discussed is to declare all the meta checks in Nixpkgs (like [this](https://github.com/NixOS/nixpkgs/blob/efcc6731f45f5e46caaf491faa705a366f105404/pkgs/stdenv/generic/check-meta.nix#L292-L360)), but implementing these checks in this tool (instead of [here](https://github.com/NixOS/nixpkgs/blob/master/pkgs/stdenv/generic/meta-types.nix)).

This way, we can also implement checks that couldn't be done in Nix, and benefit from the ratchets, such as:
- Spell checking description
- Using ratchet checks to deprecate fields or values

@adisbladis also suggested that perhaps using [jsonschema](https://json-schema.org/), though this might not be able to represent all possible checks.

PR #120, opened 2024-10-28
Clear some clippy::pedantic pings

Ben-PH:
This PR is probably one (tiny) step above scraping doc-comments and passing it through a spell-checker. I just ran clippy with `#![warn(clippy::pedantic)]`, clearing the lints and/or specifically allowing lints until it was clear. I just hope this is a net positive for you.

I noticed this project mentioned in the nixconf recording, and thought I'd have a look. This is a first-pass, and would be happy to provide more substantive PRs.

### DIFF
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -4,7 +4,6 @@ use std::{env, fs, process};
 use anyhow::Context;
 use relative_path::RelativePathBuf;
 use serde::Deserialize;
-use tempfile::Builder;
 
 use crate::nix_file::CallPackageArgumentInfo;
 use crate::problem::{
@@ -157,7 +156,7 @@ pub fn check_values(
     nix_file_store: &mut NixFileStore,
     package_names: &[String],
 ) -> validation::Result<ratchet::Nixpkgs> {
-    let work_dir = Builder::new()
+    let work_dir = tempfile::Builder::new()
         .prefix("nixpkgs-vet")
         .tempdir()
         .with_context(|| "Failed to create a working directory")?;
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -533,7 +533,11 @@ fn handle_non_by_name_attribute(
         // At this point, we completed two different checks for whether it's a `callPackage`.
         match (is_semantic_call_package, optional_syntactic_call_package) {
             // Something like `<attr> = { }`
-            (_, None) | (false, Some(_)) => {
+            (false, None)
+            // Something like `<attr> = pythonPackages.callPackage ...`
+            | (false, Some(_))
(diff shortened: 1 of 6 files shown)


src/main.rs
```diff
@@ -1,3 +1,10 @@
+// #![warn(clippy::pedantic)]
+// #![allow(clippy::uninlined_format_args)]
+// #![allow(clippy::enum_glob_use)]
+// #![allow(clippy::module_name_repetitions)]
+// #![allow(clippy::doc_markdown)]
+// #![allow(clippy::if_not_else)]
+// #![allow(clippy::ignored_unit_patterns)]
```
Ben-PH, line 7: I like to use this pattern in some of my projects. I find pedantic is often a bad mix of false positives, but having this here does two things:

a) quick-and-easy way to surface nit-picks, hiding classes of nits with fine control
b) documents where the project disagrees with the `pedantic` linter

src/eval.rs
```diff
@@ -534,11 +533,7 @@ fn handle_non_by_name_attribute(
         // At this point, we completed two different checks for whether it's a `callPackage`.
         match (is_semantic_call_package, optional_syntactic_call_package) {
             // Something like `<attr> = { }`
-            (false, None)
-            // Something like `<attr> = pythonPackages.callPackage ...`
-            | (false, Some(_))
-            // Something like `<attr> = bar` where `bar = pkgs.callPackage ...`
-            | (true, None) => {
+            (_, None) | (false, Some(_)) => {
```
philiptaron, line 536: Please revert or find another way of expressing the comments, like extracting to a function or something. Humans are bad at understanding what the `_` in these expressions actually _means_ in nixpkgs already, and the removal of the comments makes it much worse. I'd much rather have a longer function than to miss this.

src/main.rs
philiptaron: Yeah, these seem sane to me.

philiptaron (changes requested):
Looks good to me other than the one highlighted spot. Thank you for the contribution!

`cargo fmt` needs to be run as well, [as CI highlights](https://github.com/NixOS/nixpkgs-vet/actions/runs/11551592795/job/32168167952?pr=120).

src/eval.rs
Ben-PH: done: [5b4aae3](https://github.com/NixOS/nixpkgs-vet/pull/120/commits/5b4aae35ae10fc12bfed5888bba338bc34cc5fe3)

Ben-PH:
format done, requested change reversion done. Appologies for the skipped fmt.

philiptaron (approved):

Issue #107, opened 2024-09-08
nixpkgs-vet should throw error when package attribute name in by-name is prefixed with a number


Aleksanaa:
In https://github.com/NixOS/nixpkgs/pull/333281 the package attribute name is prefixed with 1 but nixpkgs-vet agreed with that. This may cause problems for us later, as nix treats `"1a"`, `1a` and `_1a` differently.

philiptaron:
I see that this _does_ actually work through both `nix build .#1fps` and `nix-build -A 1fps`, though, and also works if the attrset name is stringified.

```
nix-repl> pkgs."1fps"
«derivation /nix/store/b8d1br1rjq4i1yw8w5l71b5c26cqacrz-1fps-0.1.10.drv»

nix-repl> pkgs.1fps
error: undefined variable 'fps'

       at «string»:1:7:

            1| pkgs.1fps
             |       ^
```

Is the core reason it should be discouraged that a package like this can't be expressed as a dependency in `callPackage` argument lists?

```
nix-repl> { 1fps }: null
error: syntax error, unexpected INT

       at «string»:1:3:

            1| { 1
             |   ^
```

Aleksanaa:
> Is the core reason it should be discouraged that a package like this can't be expressed as a dependency in `callPackage` argument lists?

...Yes, in other situations we just need quotes.

infinisil:
Oh this would be nice, and should be simple to implement

infinisil:
I'll refrain from implementing this, I think this is a great first issue for perhaps a new contributor

Aleksanaa:
https://github.com/NixOS/nixpkgs-vet/blob/main/src/structure.rs#L16-L19

Ben-PH:
Feel free to assign me. I'll get started now.

As this will be the first time _navigating_ the code base, any onboarding info would be handy (Structure of code, which parts are most relevant to this, etc).

philiptaron:
> Feel free to assign me. I'll get started now.

Done! Please feel free to open a draft PR as a WIP rather than polishing until "done" -- rough drafts are just fine in these parts. 🙂 

> Any on-boarding info would be handy, like structure of the code, which parts are most relevant to this, and so forth.

1. Make a Problem or two (see the `src/problems` directory) for this issue. @Aleksanaa has highlighted a couple of possible problems, one with shards and one with packages.
2. Find the place in the code that checks package names (usually by looking for other Problems that have to do with package names)
[...]

PR #122, opened 2024-10-29
Add problem: ByNamePackagePrefixedWithNumber

Ben-PH:
Reference issue: #107 

This is first time putting together something of substance for this repo.

### DIFF
--- a/src/problem/mod.rs
+++ b/src/problem/mod.rs
@@ -34,6 +34,8 @@ pub mod npv_161;
 pub mod npv_162;
 pub mod npv_163;
 
+pub mod npv_170;
+
 #[derive(Clone, Display, EnumFrom)]
 pub enum Problem {
     /// NPV-100: attribute is not defined but it should be defined automatically
@@ -123,6 +125,9 @@ pub enum Problem {
     NewTopLevelPackageShouldBeByNameWithCustomArgument(
         npv_163::NewTopLevelPackageShouldBeByNameWithCustomArgument,
     ),
+
+    /// NPV-170: by-name package cannot be number prefixed
+    ByNamePackegPrefixedWithNumber(npv_170::ByNamePackegPrefixedWithNumber),
 }
 
 fn indent_definition(column: usize, definition: &str) -> String {
--- a/src/problem/npv_170.rs
+++ b/src/problem/npv_170.rs
@@ -1,0 +1,19 @@
+use std::fmt;
+
+use derive_new::new;
+
+#[derive(Clone, new)]
+pub struct ByNamePackegPrefixedWithNumber {
+    #[new(into)]
+    attribute_name: String,
+}
+
+impl fmt::Display for ByNamePackegPrefixedWithNumber {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let Self { attribute_name } = self;
+        write!(
+            f,
+            r#"- pkgs.{attribute_name}: "Attribute names should not be number-prefixed. It is suggestet to `"`-wrap this name"#
+        )
+    }
+}


Ben-PH:
@philiptaron I think I have the bulk of the idea here.

- copy one of the `tests/` directories. rename things to trigger the new vet-check
- creat a new module in `problem` module templated by a similar problem
- have the `Display` match the `expected` file in the new `tests/` dir-entry

...I'm just not sure about this: https://github.com/NixOS/nixpkgs-vet/actions/runs/11570703378/job/32207036558?pr=122#step:4:907

...and feedback on what the error message should actually be (I'm pretty sure the suggestion is wrong)

src/problem/npv_170.rs
```diff
@@ -0,0 +1,25 @@
...
+
+impl fmt::Display for ByNamePackegPrefixedWithNumber {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let Self {
+            package_name,
+            relative_package_dir,
+        } = self;
+        write!(
+            f,
+            r#"- {relative_package_dir}: Attribute `{package_name}` should not be number-prefixed. It is suggestet to `"`-wrap this name"#
```
willbush, line 22: > suggestet

typo

> `"`-wrap this name

I had to read that a couple of times before realizing what was meant. I think I'd prefer if it said something like "Wrap the name in quotes." However, it wasn't too clear to me when reading https://github.com/NixOS/nixpkgs/pull/333281 if wrapping in quotes is preferred over underscore prefix.

tests/by-name-numprefix/pkgs/by-name/_1/10foo/package.nix
```diff
@@ -0,0 +1,4 @@
+{ someDrv }:
+SomeDrv
+# If we caused an actual Nix failure
+# builtins.trace "This should be on stderr!" throw "This is an error!"
```
willbush, line 4: Can you delete this commented out code I think left over from the [by-name-failure test](https://github.com/Ben-PH/nixpkgs-vet/blob/107_by-name-num-prefix/tests/by-name-failure/pkgs/by-name/fo/foo/package.nix#L3)?

src/structure.rs
```diff
@@ -137,11 +140,23 @@ fn check_package(
     } else {
         let package_name_valid = PACKAGE_NAME_REGEX.is_match(&package_name);
         let result = if !package_name_valid {
-            npv_141::InvalidPackageDirectoryName::new(
-                package_name.clone(),
-                relative_package_dir.clone(),
-            )
-            .into()
+            if package_name
```
willbush, line 143: Think this if-expression can be `if package_name.starts_with(|c: char| c.is_ascii_digit())`

willbush (changes requested):
Thanks for your contributions!

willbush:
@philiptaron @infinisil forgot how where landed on how clean commit history should be for feature branches since we merge commit them into main? For example, should requested fix changed on a PR be rebased?

src/problem/npv_170.rs
Ben-PH: I'm not sure what the recommendation should be. This was a best guess. I'd happily take direction on what the warning should _explicitly_ be.

willbush:
@Ben-PH Just thought of the [readme](https://github.com/NixOS/nixpkgs-vet/blob/02db8e474807270e5611182141f764722345d319/README.md?plain=1#L68) mentioning checks. Might be worth updating that too?

src/problem/npv_170.rs
willbush: Oh looking at this again, realized they're not talking about the `pname` but the directory name for the package (which is what this checks :joy:). Can you instead only recommend the `_` prefix?

src/problem/npv_170.rs
```diff
@@ -0,0 +1,25 @@
+use std::fmt;
+
+use derive_new::new;
+use relative_path::RelativePathBuf;
+
+#[derive(Clone, new)]
+pub struct ByNamePackegPrefixedWithNumber {
```
willbush, line 7: Just noticed this typo `Packeg`

willbush (changes requested):

README.md
```diff
@@ -69,6 +69,7 @@ The following checks are performed when calling the binary:
 - `pkgs/by-name` must only contain subdirectories of the form `${shard}/${name}`, called _package directories_.
 - The `name`'s of package directories must be unique when lowercased.
 - `name` is a string only consisting of the ASCII characters `a-z`, `A-Z`, `0-9`, `-` or `_`.
+- `name` only starts with an ASCII character `a-z`, `A-Z`, `-` or `_`.
```
Aleksanaa, line 72:
```suggestion
- `name` only starts with an ASCII character `a-z`, `A-Z` or `_`.
```

It cannot start with `-`. `-` will be parsed to a negation operator.

tests/by-name-numprefix/pkgs/by-name/_1/10foo/package.nix
infinisil, line 1: Something to think about is how users are likely to run into a problem. Since users will know that the first two chars of the shard need to match the attribute name, `_1/10foo` is rather unlikely, so instead it's better to make sure it triggers on `10/10foo`.

README.md
infinisil: I don't think this needs to be done in this same PR, since it's technically a separate problem, but yeah this would be good too :)

src/structure.rs
```diff
@@ -16,8 +18,9 @@ pub const BASE_SUBPATH: &str = "pkgs/by-name";
 pub const PACKAGE_NIX_FILENAME: &str = "package.nix";
 
 lazy_static! {
-    static ref SHARD_NAME_REGEX: Regex = Regex::new(r"^[a-z0-9_-]{1,2}$").unwrap();
-    static ref PACKAGE_NAME_REGEX: Regex = Regex::new(r"^[a-zA-Z0-9_-]+$").unwrap();
+    static ref SHARD_NAME_REGEX: Regex = Regex::new(r"^((_[0-9])|([a-z][a-z0-9_-]?))$").unwrap();
+    static ref PACKAGE_NAME_REGEX: Regex =
+        Regex::new(r"^((_[0-9])|[a-zA-Z])[a-zA-Z0-9_-]*$").unwrap();
```
infinisil, line 23:
```suggestion
    static ref SHARD_NAME_REGEX: Regex = Regex::new(r"^[a-z_-][a-z0-9_-]?$").unwrap();
    static ref PACKAGE_NAME_REGEX: Regex = Regex::new(r"^[a-zA-Z_-][a-zA-Z0-9_-]*$").unwrap();
```

I think this is a bit simpler to understand

infinisil: We actually don't even need a new error type for this, because we can just update the message for when these regexes don't match here: https://github.com/NixOS/nixpkgs-vet/blob/02db8e474807270e5611182141f764722345d319/src/problem/npv_110.rs#L19 and https://github.com/NixOS/nixpkgs-vet/blob/02db8e474807270e5611182141f764722345d319/src/problem/npv_141.rs#L22

infinisil (changes requested):
Thank you! Left some comments :)

README.md
Aleksanaa: > I don't think this needs to be done in this same PR, since it's technically a separate problem, but yeah this would be good too :)

It's not allowed in the regex already.

Aleksanaa:
Hi, do you still want to work on this?

src/structure.rs
Aleksanaa: > static ref PACKAGE_NAME_REGEX: Regex = Regex::new(r"^[a-zA-Z_-][a-zA-Z0-9_-]*$").unwrap();

This allows `_abc` and even `___` as package name.

Ben-PH:
Yeah. Thanks for prompting this, had some other commitments that pulled me away from it, but ready to come back to it before the end of the week.

src/problem/mod.rs
```diff
@@ -123,6 +125,9 @@ pub enum Problem {
     NewTopLevelPackageShouldBeByNameWithCustomArgument(
         npv_163::NewTopLevelPackageShouldBeByNameWithCustomArgument,
     ),
+
+    /// NPV-170: by-name package cannot be number prefixed
+    ByNamePackegPrefixedWithNumber(npv_170::ByNamePackegPrefixedWithNumber),
```
lucasew, line 130: typo: ByNamePackegPrefixedWithNumber -> ByNamePackagePrefixedWithNumber

LSP rename could do it faster and also rename where it is used

src/structure.rs
```diff
@@ -7,7 +7,9 @@ use lazy_static::lazy_static;
 use regex::Regex;
 use relative_path::RelativePathBuf;
 
-use crate::problem::{npv_109, npv_110, npv_111, npv_140, npv_141, npv_142, npv_143, npv_144};
+use crate::problem::{
```
lucasew, line 10: was this automatically introduced by the formatter?

infinisil:
@Ben-PH Another ping, let us know if you'd like somebody else to pick this up instead :)

PR #123, opened 2024-10-30
Replace lazy_static with std::sync::LazyLock

willbush:
lazy_static is no longer needed. We can just use this from std:

https://doc.rust-lang.org/std/sync/struct.LazyLock.html

note: no difference in compile times before / after on my machine. Think compile times are dominated by clap (edit actually maybe regex-syntax in regex crate which can't be removed easily).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -302,7 +302,6 @@ dependencies = [
  "derive_more",
  "indoc",
  "itertools",
- "lazy_static",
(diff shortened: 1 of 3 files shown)


Ben-PH:
Sorry about going quiet and stalling this PR. I'm glad it got merged, and happy to make the updates.

infinisil:
No worries!

PR #124, opened 2024-10-30
Fix most clippy pedantic and nursery lint warnings

willbush:
After seeing https://github.com/NixOS/nixpkgs-vet/pull/120 I was looking at other pedantic lints and realized.. I must be pedantic because I like most of the suggestions lol. I think because this project is small having this on doesn't seem too painful. Let me know what you think!

edit: note different lint categories https://doc.rust-lang.org/nightly/clippy/

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -16,6 +16,7 @@ use std::path::{Path, PathBuf};
 use std::process::ExitCode;
 use std::{panic, thread};
 
+use crate::nix_file::NixFileStore;
 use crate::status::{ColoredStatus, Status};
 use crate::validation::Validation::Failure;
 use crate::validation::Validation::Success;
@@ -107,7 +108,7 @@ fn check_nixpkgs(nixpkgs_path: &Path) -> validation::Result<ratchet::Nixpkgs> {
         return Ok(Success(ratchet::Nixpkgs::default()));
     }
 
-    let mut nix_file_store = nix_file::Store::default();
+    let mut nix_file_store = NixFileStore::default();
     let structure = structure::check(&nixpkgs_path, &mut nix_file_store)?;
 
     // Only if we could successfully parse the structure, we do the evaluation checks
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,7 @@
 #![warn(clippy::pedantic, clippy::nursery)]
+// too many false-positives. discussion:
+// https://github.com/NixOS/nixpkgs-vet/pull/124#pullrequestreview-2405109643
+#![allow(clippy::module_name_repetitions)]
 
 mod eval;
 mod location;
--- a/src/main.rs
+++ b/src/main.rs
@@ -23,7 +23,6 @@ use std::{panic, thread};
 
 use crate::nix_file::NixFileStore;
 use crate::status::{ColoredStatus, Status};
-use crate::structure::check_structure;
 use crate::validation::Validation::Failure;
 use crate::validation::Validation::Success;
 
@@ -115,7 +114,7 @@ fn check_nixpkgs(nixpkgs_path: &Path) -> validation::Result<ratchet::Nixpkgs> {
     }
 
     let mut nix_file_store = NixFileStore::default();
-    let structure = check_structure(&nixpkgs_path, &mut nix_file_store)?;
+    let structure = structure::check(&nixpkgs_path, &mut nix_file_store)?;
 
     // Only if we could successfully parse the structure, we do the evaluation checks
     let result = structure.result_map(|package_names| {
--- a/src/main.rs
+++ b/src/main.rs
@@ -54,7 +54,7 @@ pub struct Args {
 
 fn main() -> ExitCode {
     let args = Args::parse();
-    let status: ColoredStatus = process(args.base, args.nixpkgs).into();
+    let status: ColoredStatus = process(args.base, &args.nixpkgs).into();
     eprintln!("{status}");
     status.into()
 }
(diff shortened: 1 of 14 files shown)


src/main.rs
```diff
@@ -114,8 +107,8 @@ fn check_nixpkgs(nixpkgs_path: &Path) -> validation::Result<ratchet::Nixpkgs> {
         return Ok(Success(ratchet::Nixpkgs::default()));
     }
 
-    let mut nix_file_store = NixFileStore::default();
-    let structure = check_structure(&nixpkgs_path, &mut nix_file_store)?;
+    let mut nix_file_store = nix_file::Store::default();
+    let structure = structure::check(&nixpkgs_path, &mut nix_file_store)?;
```
Ben-PH, line 111: I don't like this change. the user could end up with `Store` instead of `nix_file::Store`, and that makes for unreadable code.

If we want to use module name repetition lint, I suggest setting up pub/private namespaces, and re-exports, such that the end user must use `NixFileStore` and `structure_check`, at which point, it doesn't really make sense.

src/ratchet.rs
```diff
@@ -35,22 +35,22 @@ impl Nixpkgs {
 /// The ratchet value for a top-level package
 pub struct Package {
     /// The ratchet value for the check for non-auto-called empty arguments
-    pub manual_definition: RatchetState<ManualDefinition>,
+    pub manual_definition: State<ManualDefinition>,
```
Ben-PH, line 38: If commited to this change, make it impossible for it to be used without the word ratchet. Last thing we want is for someone to be reading this and have to put in a cognitive effort to determine which specific state is being referred to.

src/references.rs
```diff
@@ -5,15 +5,16 @@ use anyhow::Context;
 use relative_path::RelativePath;
 use rowan::ast::AstNode;
 
+use crate::nix_file::ResolvedPath;
+use crate::nix_file::Store;
```
Ben-PH, line 9:
```suggestion
use crate::nix_file;
```

cognitive overhead to determine that it is a a nix file that is being stored is a non-zero readability cost, for no discernable benefit, from what I can tell.

Ben-PH (commented):
In my opinion, the changes that I haven't commented on bring a net benefit. At worst, they make the code base no-better/no-worse, but that would require an inappropriately pessimistic view.

I do, however, strongly disagree with linting for module name repetition. With the file store for example, the changes made project the wrong intent. It is not some general store. Allowing it to be used as just a `Store`, projects the wrong notion. For this item to clearly communicate what it does, it needs to only be usable with the full name.

When I first discovered pedantic, I tried playing around with approaches to satisfy this lint in a way that makes sense. I came to the conclusion that this particular lint almost always raises false-positives.

src/problem/npv_163.rs
```diff
@@ -24,11 +24,9 @@ impl fmt::Display for NewTopLevelPackageShouldBeByNameWithCustomArgument {
...
         } = self;
         let relative_package_file = structure::relative_file_for_package(package_name);
-        let call_package_arg = if let Some(path) = call_package_path {
-            format!("./{}", path)
-        } else {
-            "...".into()
-        };
+        let call_package_arg = call_package_path
+            .as_ref()
+            .map_or_else(|| "...".into(), |path| format!("./{path}"));
```
Ben-PH, line 29: This is repeated quite a few times. I'm okay with not going DRY, but repeated code often clues you in that there's a chance to encapsulate some more business logic into the type system.

Definitely out of scope for this PR, but might be worth a follow-up issue to flag this.

src/references.rs
```diff
@@ -85,7 +83,7 @@ fn check_path(
                     )
                 })
                 .collect_vec()
-                .with_context(|| format!("Error while recursing into {}", subpath))?,
+                .with_context(|| format!("Error while recursing into {subpath}"))?,
```
Ben-PH, line 86: personally, I prefer non-inlined. Probably out of inertia/habit more than anything. The only substantive comment I have is that you can't do `{struct.field}`, meaning that sometimes you inline, sometimes you don't, bringing some inconsistencies.

Even so, I see spending too much time on this one way or the other as bike-shedding, so in any case, go ahead, and if it becomes an actual issue, we can revisit.

Ben-PH (approved):
Generally speaking I find that pedantic lints should be thought of as "opinionated nits". I like to turn them on, give it a quick pass-through, and turn them off again. I'll leave behind the things that I used to filter out the noise.

In this PR the main thing that is of substantive value is the cleaning up of arguments with respect to references, as well as passing in a ref to `Path` instead of `PathBuf` (thus removing the pathalogical constraint of "path must be on the heap and owned"). Everything else, for the most part, puts a nice polish. Over time, the cumulative gains of PRs such as this, I can see being really valuable though.

willbush:
>I like to turn them on, give it a quick pass-through, and turn them off again.

I'm not opposed to pedantic being off by default. Perhaps we could just mention it in the `CONTRIBUTING.md` as something to consider looking at. I guess I'm conflicted on one point though:

```sh
$ cargo clippy --all-targets -- -W clippy::nursery -W clippy::pedantic 2>&1 | wc -l
650
```

Running it (on `main`) produces a lot of output and it would build up overtime. The output would need to basically need to be diffed before / after PR changes which I supposed we could automate with a script. And perhaps even a GH action to show the new lints in the PR.

pros:

- might catch potential issues early
- might lower mentioning nits in PR

cons:

- might require frequent suppression
- suppression is distracting
- might be a form of unnecessary friction

willbush:
> cons:
> 
>     * might require frequent suppression
> 
>     * suppression is distracting
> 
>     * might be a form of unnecessary friction

Think I talked myself out of this. Just going to mention it in contributing.md as something one might want to checkout.

willbush:
Just going to close and perhaps recreate in another PR to cleanup the commits

PR #127, opened 2024-11-08
Lint cleanup

willbush:
Based on https://github.com/NixOS/nixpkgs-vet/pull/124, but this time fewer pedantic lint fixes and cleaner git history.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -54,7 +54,7 @@ pub struct Args {
 
 fn main() -> ExitCode {
     let args = Args::parse();
-    let status: ColoredStatus = process(args.base, args.nixpkgs).into();
+    let status: ColoredStatus = process(args.base, &args.nixpkgs).into();
     eprintln!("{status}");
     status.into()
 }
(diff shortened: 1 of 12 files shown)


willbush:
@Ben-PH feel free to review too! (Github won't let me add you as a reviewer)

src/main.rs
```diff
@@ -1,10 +1,13 @@
-// #![warn(clippy::pedantic)]
-// #![allow(clippy::uninlined_format_args)]
-// #![allow(clippy::enum_glob_use)]
-// #![allow(clippy::module_name_repetitions)]
-// #![allow(clippy::doc_markdown)]
+// Temporarily uncomment to view more pedantic or new nursery lints.
+// #![warn(clippy::pedantic, clippy::nursery)]
```
Ben-PH, line 2: :+1: 

split these into one line per warning, so that they can be selectively commented/uncommented

src/nix_file.rs
```diff
@@ -563,12 +563,7 @@ mod tests {
             (24, 11, Left("testL")),
         ];
         let expected = BTreeMap::from_iter(cases.map(|(line, column, result)| {
-            (
-                Position { line, column },
-                result
-                    .map(ToString::to_string)
-                    .map_right(|node| node.to_string()),
-            )
+            (Position { line, column }, result.map(ToString::to_string))
```
Ben-PH, line 566: :+1:

Ben-PH (commented):
I like it. 

Minor, inconsequential nit, but the map_or_else functions might be able to use `Path::display` (similar to "instead of `|s| s.to_string()`, use `str::to_string`"), but you do you.

willbush:
> Minor, inconsequential nit, but the map_or_else functions might be able to use `Path::display` (similar to "instead of `|s| s.to_string()`, use `str::to_string`"), but you do you.

Not seeing a way to do that without dropping the `./` appended to the beginning of the path.

Ben-PH:
Missed the `./` part. My mistake.

infinisil (approved):
LGTM, thanks!

PR #130, opened 2024-11-18
Automated update

infinixbot:
<details><summary>cargo changes</summary>

### cargo upgrade

```
    Checking nixpkgs-vet's dependencies
name       old req compatible latest  new req
====       ======= ========== ======  =======
clap       4.5.20  4.5.21     4.5.21  4.5.21 
serde_json 1.0.132 1.0.133    1.0.133 1.0.133
serde      1.0.214 1.0.215    1.0.215 1.0.215
   Upgrading recursive dependencies
     Locking 0 packages to latest compatible versions
note: pass `--verbose` to see 10 unchanged dependencies behind latest
note: Re-run with `--incompatible` to upgrade incompatible version requirements
note: Re-run with `--verbose` to show more dependencies
  incompatible: rowan
  latest: 14 packages

```
### cargo update

```
    Updating crates.io index
     Locking 9 packages to latest compatible versions
    Updating clap_lex v0.7.2 -> v0.7.3
    Updating errno v0.3.9 -> v0.3.10
    Updating itoa v1.0.11 -> v1.0.14
    Updating libc v0.2.162 -> v0.2.167
    Updating proc-macro2 v1.0.89 -> v1.0.92
    Updating regex-automata v0.4.8 -> v0.4.9
    Updating rustix v0.38.40 -> v0.38.41
    Updating syn v2.0.87 -> v2.0.90
    Updating unicode-ident v1.0.13 -> v1.0.14
note: pass `--verbose` to see 15 unchanged dependencies behind latest

```
### cargo outdated

```
Name   Project  Compat  Latest  Kind    Platform
----   -------  ------  ------  ----    --------
rowan  0.15.16  ---     0.16.1  Normal  ---

```
### cargo audit

```
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 689 security advisories (from /home/runner/.cargo/advisory-db)
    Updating crates.io index
    Scanning repo/Cargo.lock for vulnerabilities (91 crate dependencies)

```
</details>
<details><summary>GitHub Action updates</summary>

Bumps the actions group with 2 updates: [peter-evans/create-pull-request](https://github.com/peter-evans/create-pull-request) and [cachix/install-nix-action](https://github.com/cachix/install-nix-action).

Updates `peter-evans/create-pull-request` from 6 to 7
<details>
<summary>Release notes</summary>
[...]

PR #131, opened 2024-12-03
Fix cargo upgrade to not break by using `--incompatible`

willbush:
context: https://github.com/NixOS/nixpkgs-vet/pull/130#discussion_r1857358329

While `--incompatible` allows jumping to the next major version. it also means it doing so breaks the build then the auto PR will break every week. Remove that and we can see there are pending new major version numbers via `cargo outdated`.


philiptaron (approved):
Kind of lame, cargo.

philiptaron:
Thanks for the PR, @willbush.

PR #140, opened 2024-12-31
workflows: fix xrefcheck detection on invalid symlinks in tests dir

willbush:
Invalid symlinks are there on purpose for testing.


infinisil (approved):
Nice, thanks!

PR #142, opened 2025-01-08
introduce npv_169 to detect top level withs

lucasew:
- **add problem npv_169**
- **preparations to detect npv_169**

The plan is to detect open `with`s that have the chance of shadowing variable.

PR #143, opened 2025-01-08
Use BTreeMap to get a sorted map instead of Vec + HashMap

infinisil:
While looking at #142 I've realised that this is a good improvement

---

### DIFF
--- a/src/eval.rs
+++ b/src/eval.rs
@@ -256,8 +256,7 @@ pub fn check_values(
     );
 
     Ok(check_result.map(|elems| ratchet::Nixpkgs {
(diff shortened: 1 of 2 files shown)


lucasew:
For me this looks like a algorithmic change to another that is drop-in

My rust fu is very rusty so if its this and tests didn't break then LGTM

infinisil:
Yeah this is really just using a premade data structure that does exactly what we need instead of ad-hocing it :)

willbush (approved):

lucasew (approved):

Issue #135, opened 2024-12-14
Disallow broken string escapes


piegamesde:
I'd like to keep https://github.com/NixOS/nixpkgs/pull/365186 from regressing, so would it be possible to introduce rules that forbid these kind of strings? More broadly speaking, there are a couple of other highly deprecated features within the Nix language that nobody is using, and it might be a good idea to check for that as well.

philiptaron:
Yes, I'd welcome that.

mdaniels5757:
I wrote a rule for catching string escape issues in #211. What else is deprecated?

PR #144, opened 2025-01-08
File ratchet checks

infinisil:
This introduces the boilerplate to support ratchet (and non-ratchet) checks for all Nix files in Nixpkgs, which can be used as a basis for #142

This includes #143 

<s>Needs some code comments probably before ready for merging</s>
Added

---


infinisil:
After very minor changes and adding code comments, I think this PR is ready. While it doesn't introduce any functionality by itself, it's going to be very useful to have this for #142 and similar features. In particular https://github.com/NixOS/nixpkgs-vet/issues/135 could also be based on this.

The CI failure is unrelated, fixed by #146

src/eval.rs
```diff
@@ -156,7 +157,7 @@ pub fn check_values(
     nixpkgs_path: &Path,
     nix_file_store: &mut NixFileStore,
     package_names: &[String],
-) -> validation::Result<ratchet::Nixpkgs> {
+) -> validation::Result<BTreeMap<String, ratchet::Package>> {
```
philiptaron, line 160: Add docstring on the return value please, in particular what the key is (I _think_ it's the package name.)

src/validation.rs
```diff
@@ -75,9 +75,17 @@ impl<A> Validation<A> {
...
     /// successful. The `Problem`s of both sides are returned concatenated.
-    pub fn and<A>(self, other: Validation<A>) -> Validation<A> {
+    pub fn and_<B>(self, other: Validation<B>) -> Validation<B> {
+        self.and(other, |(), b| b)
+    }
+}
+
+impl<A> Validation<A> {
+    /// Combine two validations, both of which need to be successful for the return value to be
+    /// successful. The `Problem`s of both sides are returned concatenated.
```
philiptaron, line 85: This is great and much needed.

src/ratchet.rs
```diff
@@ -57,6 +62,19 @@ impl Package {
...
+pub struct File {}
+
+impl File {
+    /// Validates the ratchet checks for a top-level package
+    pub fn compare(
+        _name: &RelativePath,
+        _optional_from: Option<&Self>,
+        _to: &Self,
+    ) -> Validation<()> {
+        Success(())
```
philiptaron, line 74: This is where #142 will go, right?

src/ratchet.rs
```diff
@@ -27,6 +29,9 @@ impl Nixpkgs {
                 .into_iter()
                 .map(|(name, pkg)| Package::compare(&name, from.packages.get(&name), &pkg)),
         )
+        .and_(validation::sequence_(to.files.into_iter().map(
+            |(name, file)| File::compare(&name, from.files.get(&name), &file),
+        )))
```
philiptaron, line 34: I see: after all the package checks, check all the files directly.

src/main.rs
```diff
@@ -113,20 +115,30 @@ fn check_nixpkgs(nixpkgs_path: &Path) -> validation::Result<ratchet::Nixpkgs> {
...
+    };
+
+    let file_result = files::check_files(&nixpkgs_path, &mut nix_file_store)?;
 
-    Ok(result)
+    Ok(
+        package_result.and(file_result, |packages, files| ratchet::Nixpkgs {
+            packages,
+            files,
+        }),
```
philiptaron, line 140: Future: could use a derived `From` here like the errors do. But this is quite clear.

src/files.rs
```diff
@@ -0,0 +1,77 @@
...
+use crate::{nix_file, ratchet, structure, validation};
+
+/// Runs check on all Nix files, returning a ratchet result for each
+pub fn check_files(
+    nixpkgs_path: &Path,
+    nix_file_store: &mut NixFileStore,
+) -> validation::Result<BTreeMap<RelativePathBuf, ratchet::File>> {
+    process_nix_files(nixpkgs_path, nix_file_store, |_nix_file| {
+        // Noop for now, only boilerplate to make it easier to add future file-based checks
+        Ok(Success(ratchet::File {}))
```
philiptaron, line 18: Or is this the place that #142 will go?

src/files.rs
```diff
@@ -0,0 +1,77 @@
...
+    nixpkgs_path: &Path,
+    nix_file_store: &mut NixFileStore,
+    f: impl Fn(&nix_file::NixFile) -> validation::Result<ratchet::File>,
+) -> validation::Result<BTreeMap<RelativePathBuf, ratchet::File>> {
+    // Get all Nix files
+    let files = {
+        let mut files = vec![];
+        collect_nix_files(nixpkgs_path, &RelativePathBuf::new(), &mut files)?;
+        files
+    };
```
philiptaron, line 34: I don't think I've seen this pattern before. I like it!

src/files.rs
```diff
@@ -0,0 +1,77 @@
...
+    let results = files
+        .into_iter()
+        .map(|path| {
+            // Get the (optionally-cached) parsed Nix file
+            let nix_file = nix_file_store.get(&path.to_path(nixpkgs_path))?;
+            let result = f(nix_file)?;
+            let val = result.map(|ratchet| (path, ratchet));
+            Ok::<_, anyhow::Error>(val)
+        })
+        .collect_vec()?;
```
philiptaron, line 45: It's possible to use [itertools::process_results](https://docs.rs/itertools/latest/itertools/fn.process_results.html) to avoid collecting into a vec before turning into a BtreeMap, but this is totally fine.

philiptaron (approved):
No blocking comments, just a few clarifying questions. Thanks for your patience in waiting for a review.

src/ratchet.rs
infinisil: No logic, it only needs to do the same as for `Package`: https://github.com/NixOS/nixpkgs-vet/blob/3a7a1b8b0e64c94c66f5dfd4f78fb5379c311d22/src/ratchet.rs#L44-L57

src/files.rs
infinisil: This is where the logic goes yes :)

src/files.rs
infinisil: Neither have I! Was wondering how acceptable it was, but I guess you're fine with it :smile:

src/files.rs
infinisil: Doesn't seem to quite work because we only have `validation::sequence` to combine a sequence of `Validation`'s, which is based on `Vec`s. But glad to know about `process_results`, it effectively does the same as `collect_vec`, I'll make use of it!

infinisil: Oh actually no not really, `collect_vec` is only there to be a shorthand, but with `process_results` it would be `process_results(|x| x.collect_vec())`. I'll keep this the same for now :)

src/eval.rs
infinisil: Done :)

philiptaron (approved):

PR #145, opened 2025-01-08
Introduce bogus ratchet check ensuring that no new Nix string files are introduced

infinisil:
Based on top of #144 as an example of how something like this can be done for #142 

---


philiptaron (approved):
Thanks, this is really helpful for understanding #144. Somehow did not see it until just now.

PR #146, opened 2025-01-14
Make xrefcheck ignore intentionally invalid symlinks

infinisil:
Fixes https://github.com/NixOS/nixpkgs-vet/actions/runs/12776335415/job/35614770890?pr=144 by adding a default `xrefcheck.yaml` file and then adding explicit ignores for the intentionally invalid symlinks.

### DIFF
--- a/xrefcheck.yaml
+++ b/xrefcheck.yaml
@@ -1,0 +1,109 @@
+# Generated using `xrefcheck dump-config -t GitHub` at version 0.3.1
+
+# Exclusion parameters.
+exclusions:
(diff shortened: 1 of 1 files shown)


infinisil:
Oh, just realising that https://github.com/NixOS/nixpkgs-vet/pull/140 was supposed to fix that too, not sure why that doesn't work

infinisil:
Nevermind! https://github.com/NixOS/nixpkgs-vet/pull/144#issuecomment-2591118861

PR #158, opened 2025-05-10
feat: nixpkgs-vet --version

dtomvan:
This PR simply adds the `clap`-flag `version`, so that `-V` or `--version` can be passed to get the current version of `nixpkgs-vet`.

Why: to make `versionCheckHook` work. Some effort is being made right now to package `nixpkgs-vet` inside of `nixpkgs` ([here](https://github.com/NixOS/nixpkgs/pull/405853)) and `versionCheckHook` is a good practice to allow r-ryantm to test its updates, and a way to make sure that a build only succeeds when the output binary has at least _some_ basic functionality.

PR #160, opened 2025-05-12
.github: run CI with `max-jobs` set to 1 in order to avoid OOM

philiptaron:
As suggested by @dtomvan in [this comment](https://github.com/NixOS/nixpkgs-vet/pull/158#issuecomment-2869670960), try to avoid OOM through `max-jobs = 1`.

I'm planning on self-merging if it goes through CI successfully, because I regressed these checks trying to fix the update process.


philiptaron:
Turns out I can't self-merge. @willbush, @infinisil -- mind reviewing?

willbush (approved):

Issue #31, opened 2024-03-28
Support other package sets


infinisil:
We want to use the `pkgs/by-name` convention for other package sets, e.g. https://github.com/NixOS/nixpkgs/pull/299347.

Here's a rough idea how this could look like:

## Package set config

Define some Nix config file to determine which package sets exist:

```nix
# pkgs/top-level/by-name-sets-config.nix
{
  packageSets = {
    toplevel = {
      attrPath = [ ];
      path = ./pkgs/by-name;
      shardLength = 2;
    };
    mySet = {
      # Any attrpath that will contain the attributes
      attrPath = [ "mySet" ];
      path = ./pkgs/mySet/by-name;
      # No sharding
      shardLength = null;
    };
  };
}
```

## Read the packages in Nix

Create a Nix file to automatically call all of them like the current [`by-name-overlay.nix`](https://github.com/NixOS/nixpkgs/blob/4c731c3c34b5e79c68c6210c827097cb284b80ab/pkgs/top-level/by-name-overlay.nix):

```nix
# pkgs/top-level/by-name-sets.nix
{
  toplevel = callPackage: {
    hello = <...>;
  };
  mySet = callPackage: {
    hi = <...>;
  };
}
```

For each package set declared, define it by importing `by-name-sets.nix`, like

```nix
mySet = lib.makeScope pkgs.newScope (self:
  (import ./pkgs/top-level/by-name-sets.nix).mySet self.callPackage
) 
```

## Check the packages with the tool

Adjust this tool to read a `by-name-sets-config.nix` file for determining what it needs to check and how.

infinisil:
Had a chat today with @Dmills27, who's interested in contributing. I took some notes to break down this task:

1. Make [`by-name-overlay.nix`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/by-name-overlay.nix) be synced with the tool
   - Move `by-name-overlay.nix` into nixpkgs-check-by-name
   - Make sure it can accept a config argument in the future
   - Distribute this file in the releases of nixpkgs-check-by-name
     - Not necessary, Github always distributes the source tarball
   - In Nixpkgs, the [`update-pinned-tool.sh`](https://github.com/NixOS/nixpkgs/blob/051f920625ab5aabe37c920346e3e69d7d34400e/pkgs/test/check-by-name/update-pinned-tool.sh) should be updated to also fetch the Nix file from the release.
2. Add a config file with a configurable prefix length
[...]

PR #180, opened 2025-10-17
[WIP] Multiple by-name directory support

mdaniels5757:
This works in tests only.

Known issues:
1. This is my first time really programming in Rust, and, um, that's probably evident.
2. Memory usage during eval is absurd at the moment (45GB+ per eval), and then the stack overflows/infinite recursion.
1. This takes about 2.5x the amount of time it did before. I need to profile.

One thing that may or may not be an issue, is that I mostly did not follow the instructions @infinisil so kindly left on #31. Not sure which of those should or should not be implemented.

Feedback is welcome! I would ask you please refrain for now from e.g. pointing out commented-out code and the like. (I'm well aware this needs cleanup.) But more function- or design- focused feedback, or feedback related to the points above, is definitely welcome (and probably needed :) )!


mdaniels5757:
Made some changes, but I'm now getting issues with `tclPackages.tclcurl: This attribute is not defined but it should be defined automatically as pkgs/development/tcl-modules/by-name/tc/tclcurl/package.nix` (where it absolutely is). (I'm also getting, `tclPackages.bwidget: This attribute is not defined but it should be defined automatically as pkgs/development/tcl-modules/by-name/tc/bwidget/package.nix`, with the incorrect shard, but I can fix that, I think.).

Some thoughts about what could cause the former would be appreciated.

infinisil (changes requested):
Finally got around to taking a closer look at this, and overall this is already looking very solid, thanks a lot!

For now I'll just comment for the biggest issue I've noticed, I can do another review when that's addressed.

PR #181, opened 2025-10-21
Remove support for by-name overrides

wolfgangwalther:
This is for when NixOS/nixpkgs#454525 is done.

It should be fine to implement that nixpkgs-side check only on unstable, but still remove it from nixpkgs-vet for both branches. 25.05 will not be alive much longer and it's *very* unlikely that backports would introduce bad code that way. And even if they do - it doesn't really matter in the bigger picture.

I'm hoping this would make #180 simpler. cc @mdaniels5757


wolfgangwalther:
Just rebased after merging #183.

PR #189, opened 2026-01-26
Remove support for by-name overrides

mdaniels5757:
Resurrection of #181.

Needs NixOS/nixpkgs#483820.


mdaniels5757:
Thinking: since we use the same nixpkgs-vet version for stable and unstable, this probably needs to wait until June. :(

philiptaron:
Or we need to make the tool aware of branch status, which we absolutely can do.

mdaniels5757:
> Or we need to make the tool aware of branch status, which we absolutely can do.

Sure, but my goal in doing this is (in significant part) to make #180 much simpler to implement. If all of the code is still present (even if some of it is feature flagged) that doesn't really help me, I think? But maybe I don't fully understand what you mean?

qweered:
Are there any blockers left? We are waiting until 25.11 EOL, right?

mdaniels5757:
Correct, I intend to rebase and undraft in 2 weeks or so.

MattSturgeon:
> We are waiting until 25.11 EOL, right?

> marked this pull request as ready for review 

Looking at [oldestSupportedRelease](https://github.com/NixOS/nixpkgs/blob/535db66ef71b918aea52078551a9a594f1e36b37/lib/trivial.nix#L475-L477) and [channels.nix](https://github.com/NixOS/infra/blob/main/channels.nix#L58-L62), we're still supporting 25.11 (as deprecated) for now. Are we no longer blocked by that or do you want to land this _in preparation_ for bumping `ci/pinned.json` later, once 25.11 is unmaintained?

src/problem/mod.rs
```diff
@@ -200,9 +152,7 @@ impl Problem {
             Self::NixFileIsExecutableWithoutShebang(..) => "NPV-145",
             Self::NixFileHasShebangButNotExecutable(..) => "NPV-146",
             Self::TopLevelPackageMovedOutOfByName(..) => "NPV-160",
-            Self::TopLevelPackageMovedOutOfByNameWithCustomArguments(..) => "NPV-161",
```
MattSturgeon, line 203: Q: why remove this one?

src/problem/mod.rs
```diff
@@ -200,9 +152,7 @@ impl Problem {
...
             Self::NewTopLevelPackageShouldBeByName(..) => "NPV-162",
-            Self::NewTopLevelPackageShouldBeByNameWithCustomArgument(..) => "NPV-163",
```
MattSturgeon, line 205: Q: why remove this one?

src/problem/mod.rs
```diff
@@ -173,13 +132,6 @@ impl Problem {
         match self {
             Self::ByNameUndefinedAttribute(..) => "NPV-100",
             Self::ByNameNonDerivation(..) => "NPV-101",
-            Self::ByNameInternalCallPackageUsed(..) => "NPV-102",
```
MattSturgeon, line 176: Removing this implies we will also remove `_internalCallByNamePackageFile` from nixpkgs: https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/by-name-overlay.nix#L50-L56

src/problem/mod.rs
```diff
@@ -173,13 +132,6 @@ impl Problem {
...
-            Self::ByNameCannotDetermineAttributeLocation(..) => "NPV-103",
-            Self::ByNameOverrideOfNonSyntacticCallPackage(..) => "NPV-104",
-            Self::ByNameOverrideOfNonTopLevelPackage(..) => "NPV-105",
-            Self::ByNameOverrideContainsWrongCallPackagePath(..) => "NPV-106",
-            Self::ByNameOverrideContainsEmptyArgument(..) => "NPV-107",
-            Self::ByNameOverrideContainsEmptyPath(..) => "NPV-108",
```
MattSturgeon, line 182: These all relate to overrides, SGTM.

src/eval.rs
MattSturgeon, line 1: Should we add a changeset file for knope to use in the changelog?

MattSturgeon (commented):
I can't meaningfully review this, as I'm not really familiar enough with the codebase to know if a) everything relevant was removed (nothing was missed) and b) everything removed was relevant (nothing unrelated was removed).

From a read over the diff, it looks reasonable.

src/problem/mod.rs
mdaniels5757: That's the plan!

src/problem/mod.rs
mdaniels5757: Since we only _really_ needed the empty_arg field for https://github.com/NixOS/nixpkgs-vet/commit/1908f81552c55c602775d7f4ecb03fcb46aeebf6#diff-84c46c1047bea02643d764007dc4ae2cb6e93767ba92faa9583659cb88b49bf4L452-L458, we can now remove it. That means we need to get rid of NPV-161 and NPV-163, using NPV-160 and NPV-162 instead.

src/problem/mod.rs
mdaniels5757: Responding to the other one.

src/eval.rs
mdaniels5757: Done!

philiptaron:
https://github.com/mdaniels5757/nixpkgs-vet/pull/17

Few more things I noticed reviewing the PR. LGTM btw.

philiptaron (approved):

philiptaron:
I updated the Wiki for this also

PR #197, opened 2026-02-16
Update rnix 0.12.0 to 0.14.0 with new path lints

philiptaron:
## Summary

- Update `rnix` from 0.12.0 to 0.14.0 and `rowan` from 0.15.17 to 0.16.1
- Leverage rnix 0.13+'s type-level path variants (`PathAbs`, `PathRel`, `PathHome`, `PathSearch`) instead of string-prefix hacks
- Add **NPV-127**: absolute path expressions (e.g. `/foo`) are not allowed in nixpkgs
- Add **NPV-128**: home-relative path expressions (e.g. `~/foo`) are not allowed in nixpkgs
- Absolute paths now get a specific error instead of falling through to the generic "cannot be resolved" (NPV-124)

## Test plan

- [x] `cargo build` compiles successfully
- [x] `cargo test` — all 6 tests pass
- [x] `tests/ref-absolute` updated to expect NPV-127 instead of NPV-124
- [x] `tests/multiple-failures` updated for the new absolute path message
- [x] `tests/ref-home-path` added as new test case for NPV-128
- [x] `tests/ref-nix-path` unchanged (search path detection still works)

### DIFF
--- a/src/references.rs
+++ b/src/references.rs
@@ -170,15 +170,13 @@ fn check_nix_file(
                     text,
                 )
                 .into(),
-                ResolvedPath::HomeRelativePath => {
-                    npv_128::NixFileContainsHomeRelativePath::new(
-                        relative_package_dir,
-                        subpath,
-                        line,
-                        text,
-                    )
-                    .into()
-                }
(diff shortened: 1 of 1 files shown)


mdaniels5757 (approved):
LGTM!

Issue #110, opened 2024-09-14
Disallow executable bit on .nix files


patka-123:
I came across [a nixpkgs PR](https://github.com/NixOS/nixpkgs/pull/341771) that removes the executable bit from some .nix files. This can be prevented by a check in CI.

Would this project (nixpkgs-vet) be appropriate for adding such a check, or would it fit better in a standalone loose action?

If we decide where it should go I'd like to take a stab at it.

infinisil:
Yeah this would be a pretty good fit for such a check! In particular it would fit around here: https://github.com/NixOS/nixpkgs-vet/blob/0d26da8aa2bbc5d74461ca72391d6260ea51d028/src/structure.rs#L170-L176

I think for this case it might be fine to just add a simple additional check and error, but in theory this _could_ break master when a PR introduces an executable file but already went through CI before this change. So whenever strictness of checks is increased, a [ratchet check](https://github.com/NixOS/nixpkgs-vet/tree/main?tab=readme-ov-file#ratchet-checks) would be most ideal to prevent such situations.

If you do want to go for the full ratchet check, you'd have to extend [`ratchet::Package`](https://github.com/NixOS/nixpkgs-vet/blob/0d26da8aa2bbc5d74461ca72391d6260ea51d028/src/ratchet.rs#L34-L40) with a something like a `RatchetState<Executable>`, change [`check_package`](https://github.com/NixOS/nixpkgs-vet/blob/0d26da8aa2bbc5d74461ca72391d6260ea51d028/src/structure.rs#L119-L125) to return `validation::Result<HashMap<String, RatchetState<Executable>>>`, and then incorporate those `RatchetState<Executable>` values into the construction of `ratchet::Package` in [`eval.rs`](https://github.com/NixOS/nixpkgs-vet/blob/0d26da8aa2bbc5d74461ca72391d6260ea51d028/src/eval.rs#L154) (or a refactoring to move the construction out of that file).

While this would be neat, no offense taken if you don't go for that, a non-ratchet check is fine too (and perhaps somebody from @NixOS/nixpkgs-vet would be interested in making it a ratchet one) :)

infinisil:
> Does it really break master though? A nix file with an executable bit falling through the cracks doesn't seem like the end of the world to me.

Scenario:
- There's a PR that adds an executable file, CI is successful because there's no check preventing that just yet
- We implement a check to ensure there are no executables, but it only runs for PRs that get updated
[...]

PR #199, opened 2026-02-16
On .nix files, disallow executable bit without shebang (NPV-145) and shebang without executable bit (NPV-146)

philiptaron:
## Summary

- Adds NPV-145: detects `.nix` files with the executable bit set that lack a shebang (`#!`) line
- This is a non-ratchet (hard error) check applied to all `.nix` files in nixpkgs
- Wiki updated with [NPV-145](https://github.com/NixOS/nixpkgs-vet/wiki/NPV-145) page

Closes NixOS/nixpkgs-vet#110

## Test plan

- [x] New test case `nix-file-executable` with an executable `package.nix` (no shebang) verifies the error is reported
- [x] All existing tests continue to pass
- [x] `cargo fmt -- --check` passes

### DIFF
--- a/src/files.rs
+++ b/src/files.rs
@@ -1,9 +1,13 @@
 use relative_path::RelativePath;
 use relative_path::RelativePathBuf;
 use std::collections::BTreeMap;
+use std::fs;
+use std::io::Read;
+use std::os::unix::fs::PermissionsExt;
 use std::path::Path;
 
 use crate::nix_file::NixFileStore;
+use crate::problem::{npv_145, npv_146};
 use crate::validation::ResultIteratorExt;
 use crate::validation::Validation::Success;
 use crate::{nix_file, ratchet, structure, validation};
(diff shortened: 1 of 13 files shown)


src/files.rs
```diff
@@ -50,6 +54,30 @@ fn process_nix_files(
     }))
 }
 
+/// Check that a Nix file is not executable, unless it has a shebang (`#!`) line.
+fn check_not_executable(
```
mdaniels5757: This function only does half of what was specified in https://github.com/NixOS/nixpkgs-vet/issues/110#issuecomment-2959177413: it disallows +x on all files without a shebang, but does not force +x on all files with a shebang.

Is that intentional?

Also, I don't have a much better suggestion for the name, but I consider the current one a bit misleading. `check_executable_iff_shebang` (if the implementation is changed to match the issue comment) is a bit long but more accurate. `check_exe_iff_shebang`? Or if you decide to keep the implementation as-is, `check_executable_implies_shebang`? Hopefully you have better ideas than mine... :)

mdaniels5757 (commented):
Mostly looks good to me.

Ideally I think there would be another test case or 2 covering the "+x with shebang" and/or "-x with shebang" cases. (We have no shortage of the "-x without shebang" case, of course :) )

philiptaron:
I've added NPV-146 and an accompanying nixpkgs PR for it

mdaniels5757 (approved):
Looks good to me, but this will need #491208 and #491190 to make it to the `nixpkgs-unstable` channel, then the automated update workflow to be run, then a rebase.

PR #244, opened 2026-05-12
Automate releases with knope

### LINKED ISSUE #38: Version and changelog
We should have a changelog and decide the versioning scheme along with it. These are some thoughts regarding that, prompted via https://github.com/NixOS/nixpkgs-check-by-name/pull/37.

I'm not considering changelog tools that use the Git log, because I don't think the Git log is valuable for users. I'd like to require each PR to submit such a changelog entry file (with an option to explicitly declare the PR as not changelog-worthy) (first seen [here](https://news.ycombinator.com/item?id=34525265)), and then using a tool to generate the changelog when it's time for a release:
- [changelog-d](https://codeberg.org/fgaz/changelog-d) (used by Nix [since recently](https://github.com/NixOS/nix/pull/9393))
- [changelogd](https://changelogd.readthedocs.io) (what I first thought Nix used..)
- [Towncrier](https://towncrier.readthedocs.io/) (looks very promising)
- [Scriv](https://scriv.readthedocs.io/) (not as feature-complete, but does have some built-in support for GitHub releases)
- [Changie](https://github.com/miniscruff/changie) (looks very promising)
- (others?)

Regarding versions, there's [various versioning schemes](https://lobste.rs/s/mkaigy/solover_is_simple_expressive_versioning#c_2ruerf), but I'm liking [EffVer](https://jacobtomlinson.dev/effver/) the most, and it matches the existing versions 0.1.0 and since [recently](https://github.com/NixOS/nixpkgs-check-by-name/pull/36) 0.1.1.

I would like the automated weekly update PRs to also take care of bumping the release and changelog automatically if there's been any changes. To figure out the next version, all change entries should be tagged with the level of effort required to update (one of macro/meso/micro). Then the next version can be generated by increasing the previous one by that level. Though it's not perfect, because 100 micro changes could amount to a meso/macro change. So being able to adjust the automatically generated version number would be good too (also for an eventual 1.0.0 release).

Note that we should take some precaution to not miss PRs that might get merged _during_ which the release/update PR is open, probably by making sure that a commit gets tagged instead of the merged commit.

mdaniels5757:
Things done:

- [x] Tested on my fork
- [x] NixOS/org#252 is completed

Fixlist:
- Closes: #38
- Closes: #46


MattSturgeon (approved):
I've been putting off reviewing this as I don't have the energy to go through line-by-line. But if you're confident in it, I'm happy to rubber-stamp an approval.

I'm excited to see what the knope release-workflow looks like in practice; managing prospective releases via PRs seems like a good approach.

mdaniels5757:
It worked well on my fork, so modulo authentication (which I just double-checked) it should work well. Let's see!

PR #249, opened 2026-06-01
Automated update

infinixbot:
Running /nix/store/l9rc61srlrkwhppxl1idvbyxaa0xb8mc-cargo
<details><summary>cargo changes</summary>

### cargo upgrade

```
    Checking nixpkgs-vet's dependencies
note: Re-run with `--verbose` to show more dependencies
  latest: 18 packages

```
### cargo update

```
    Updating crates.io index
     Locking 4 packages to latest Rust 1.95.0 compatible versions
    Updating bitflags v2.11.1 -> v2.12.1
    Updating log v0.4.30 -> v0.4.32
    Updating memchr v2.8.0 -> v2.8.1
    Updating unicode-segmentation v1.13.2 -> v1.13.3

```
### cargo outdated

```
All dependencies are up to date, yay!

```
### cargo audit

```
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 1120 security advisories (from /home/runner/.cargo/advisory-db)
    Updating crates.io index
    Scanning repo/Cargo.lock for vulnerabilities (101 crate dependencies)

```
</details>
Running /nix/store/q78qkrbvinxiyx0wxki4gxlhpbn7gxb4-update-npins
<details><summary>npins changes</summary>

```
[treefmt-nix] Changes:
-    revision: 790751ff7fd3801feeaf96d7dc416a8d581265ba
+    revision: db947814a175b7ca6ded66e21383d938df01c227
-    url: https://github.com/numtide/treefmt-nix/archive/790751ff7fd3801feeaf96d7dc416a8d581265ba.tar.gz
+    url: https://github.com/numtide/treefmt-nix/archive/db947814a175b7ca6ded66e21383d938df01c227.tar.gz
-    hash: sha256-pc20NRoMdiar8oPQceQT47UUZMBTiMdUuWrYu2obUP0=
+    hash: sha256-eynAfOmbmxJnkp7YewvCEbShNnnYJ9gLLqkzsYtBPeM=
[nixpkgs] Changes:
-    url: https://releases.nixos.org/nixpkgs/nixpkgs-26.05pre1004168.d849bb215dcd/nixexprs.tar.xz
+    url: https://releases.nixos.org/nixpkgs/nixpkgs-26.11pre1009182.ffa10e26ae11/nixexprs.tar.xz
-    hash: sha256-rOwHOh5nbIqhJxISJpvdOpvvE7gHD9Hmu229/RVvnjE=
[...]

PR #251, opened 2026-06-04
ci: replace update-github-actions script with hosted Dependabot

mdaniels5757:
Fixes the error observed in #249.


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
```
MattSturgeon, line 3: Future idea: we could consider using dependabot for flake updates\* and cargo updates.

\*I don't think dependabot supports npins yet.

MattSturgeon (approved):

philiptaron (approved):

PR #257, opened 2026-06-13
refactor: npins → flake

MattSturgeon:
Use an empty flake to manage our nix inputs instead of npins and replace our custom update script with GitHub-hosted dependabot.

- `flake.nix` and `flake.lock` now manage inputs
- dependabot now updates flake inputs
- dependabot now also updates cargo inputs
- we already used dependabot for GHA actions

In a follow-up PR we could extend the new flake to also produce outputs, similar to what we do in nixfmt:
https://github.com/NixOS/nixfmt/blob/f5193ad037aebee328ffc0976f978ee91e886011/flake.nix#L32-L61

---

As I mentioned in https://github.com/NixOS/nixpkgs/pull/431124#pullrequestreview-3088020976 and more recently https://github.com/NixOS/nixfmt/pull/416, I'd like to do something like this in Nixpkgs' CI too, to replace its `pinned.json`. Any thoughts on that?

I assume in Nixpkgs, we'd want to read the lockfile manually, rather than depending on flake-compat. Something like:
```nix
lock = builtins.fromJSON (builtins.readFile ./flake.lock); 
inputNodes = builtins.mapAttrs (_: nodeName: lock.nodes.${nodeName}) lock.nodes.${lock.root}.inputs;
inputs = builtins.mapAttrs (_: node: builtins.fetchTarball {/*...*/) inputNodes;
```

or similar to Nixvim's https://github.com/nix-community/nixvim/blob/f2029d9a26266eb67f46b0c79bd0a3713839a57a/nixpkgs.nix


default.nix
```diff
@@ -82,80 +79,6 @@ let
...
-
-              echo "### cargo outdated"
-              printf "\n\`\`\`\n"
-              cargo outdated --manifest-path "$1/Cargo.toml" 2>&1
-              printf "\n\`\`\`\n"
-
-              echo "### cargo audit"
-              printf "\n\`\`\`\n"
-              cargo audit --file "$1/Cargo.lock" 2>&1
-              printf "\n\`\`\`\n"
```
MattSturgeon, line 139: Q: how much of this is done by Dependabot? How much of it is actually important?

Do we want to do any of this in a dedicated workflow, e.g. an `on: pull_request: paths: Cargo.lock` workflow?

.github/workflows/update.yml
MattSturgeon, line 1: Running the `update.yml` workflow (without creating the update PR) is a 'required status check'.

Therefore, if merging this PR, we'll need to update the ruleset first.

default.nix
mdaniels5757: * `cargo update` is done by default.
* `cargo upgrade` is done if you set the [`versioning-strategy`](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#versioning-strategy--) option to `increase`.
* `cargo outdated` shouldn't be needed.
* `cargo audit` should be handled by Dependabot Security Alerts, which I've just enabled.

.github/dependabot.yml
```diff
@@ -9,3 +9,17 @@ updates:
       actions:
         patterns:
           - "*"
+
+  - package-ecosystem: "nix"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
mdaniels5757, line 16:
```suggestion
      interval: "weekly"
      day: "monday"
```

.github/dependabot.yml
```diff
@@ -9,3 +9,17 @@ updates:
...
+
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
mdaniels5757, line 22:
```suggestion
      interval: "weekly"
      day: "monday"
```

.github/dependabot.yml
```diff
@@ -9,3 +9,17 @@ updates:
...
+    groups:
```
mdaniels5757, line 25: This replaces `cargo upgrade`.

```suggestion
    versioning-strategy: "increase"
    groups:
```

mdaniels5757 (approved):

MattSturgeon:
> Running the `update.yml` workflow (without creating the update PR) is a 'required status check'.
> 
> Therefore, if merging this PR, we'll need to update the ruleset first.

Looks like anyone in nixpkgs-ci can update https://github.com/NixOS/nixpkgs-vet/settings/rules/541493#_r_19_-label

I'm happy to do that now if we're ready to merge?

mdaniels5757:
Refresh the page -- I already did :)

MattSturgeon:
Hm, dependabot is [complaining](https://github.com/NixOS/nixpkgs-vet/runs/81150922174):

> The property '#/updates/2/versioning-strategy' value "increase" did not match one of the following values: lockfile-only, auto

But I can clearly see `increase` documented in [`versioning-strategy`](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#versioning-strategy--)

Seems `increase` would be the default anyway, if dependabot considers `nixpkgs-vet` an "app" rather than a "library":

> For apps, always increase the minimum version requirement to match the new version. The `increase` strategy.

PR #262, opened 2026-06-13
ci: fix knope

mdaniels5757:
Actually fix the problem described in #261, following the advice of the note in
https://knope.tech/recipes/1-preview-releases-with-pull-requests/#prepare_releaseyaml.

Also fix my syntax highlighting/editor for the GitHub actions workflow by
changing the format of the if expression. No substantive change.

Tested in my fork.

### DIFF
--- a/knope.toml
+++ b/knope.toml
@@ -35,6 +35,10 @@ command = "git switch -c release"
 type = "PrepareRelease"
 allow_empty = true
 
+# push-draft-release workflow
+[[workflows]]
+name = "push-draft-release"
+
 [[workflows.steps]]
 type = "Command"
 command = "git commit -m \"chore: prepare release $version\""
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,10 +42,10 @@ jobs:
         run: nix-build -A ci
 
   prepare-release:
-    if: "${{
-        github.event_name == 'push' &&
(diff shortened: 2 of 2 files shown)


knope.toml
```diff
@@ -35,6 +35,10 @@ command = "git switch -c release"
 type = "PrepareRelease"
 allow_empty = true
```
MattSturgeon, line 36: Does this mean we should drop:

```suggestion
```

MattSturgeon (approved):

knope.toml
mdaniels5757: No, because then the step will fail.

MattSturgeon: Ah, I figured the `if` condition on the job fully intersected with the "empty" case we're allowing

mdaniels5757: We don't have any if condition on this GitHub Action step (and this Knope step is the one that bumps the version).

MattSturgeon: We have:
```yaml
     # Skip merges from releases 
     if: >- 
       github.event_name == 'push' && 
       !contains(github.event.head_commit.message, 'chore: prepare release')
```

...on the prepare-release job (which this config applies to IIUC). However looking closer, that is filtering out merges from release PRs, not merges from "empty" PRs, so you're right that we still need `allow_empty`.

PR #265, opened 2026-06-15
Ignore ineffective package attrs in `strictDeps` & `__structuredAttrs` checks

### LINKED ISSUE #263: Check if attributes are in the derivation or not
Based upon my brief reading, nixpkgs-vet will accept new derivations that set `passthru = { strictDeps = true; __structuredAttrs = true; }`. 

This will not actually turn on either one of the features. 

The fix would be to check `<package>.drvAttrs.{strictDeps, __structuredAttrs}`

RossSmyth:
Closes #263 

For `strictDeps` and `structuredAttrs` to be effectively set, they must go through `stdenv`'s machinery. The direct way of checking that would be something like 
```
> nix derivation show -f. hello | jq '((.[].env.strictDeps // "0") == "1") or (.[].env.__json | fromjson | .strictDeps)'
```
and
```
> nix derivation show -f. hello | jq '.[].env | has("__json")'
```
Since I don't want to change everything to do that, this is the next best thing. Without these checks people can write:
```nix
stdenv.mkDerivation {
  passthru = {
    strictDeps = true;
    __structuredAttrs = true;
  };
}
```
And the check will pass. Alternatively:
```nix
package-final = package // {
  strictDeps = true;
  __structuredAttrs = true;
};
```
and tests will pass. ~~What this cannot guard against is people writing:~~
```nix
package-final = package // {
  drvAttrs = package.drvAttrs // {
    strictDeps = true;
    __structuredAttrs = true;
  };
}
```
~~and vet passing.~~ But hopefully this will look weird enough that it will be caught in review. But I would've thought the same for the latter ones and they weren't immediately caught. 

Codebase could use some documentation on what arguments do and such, as it took me a bit and still don't completely get it. But tests pass 😎 👍

### DIFF
--- a/src/eval.nix
+++ b/src/eval.nix
@@ -58,6 +58,10 @@ let
         { NonAttributeSet = null; }
       else
         let
+          # Not all attrs (or even all derivations) have an `overrideAttrs`,
+          # only those constructed by something based on `stdenv.mkDerivation`
+          overrideValue = value.overrideAttrs or (_: value);
+
           # Disallows people getting around actually setting `strictDeps`
           # and `__structuredAttrs` by doing something like:
           # {
@@ -67,20 +71,14 @@ let
           # package = package-final // {
           #   __structuredAttrs = true;
           # };
-          cleanPackage =
-            if value ? overrideAttrs then
-              value.overrideAttrs (
-                _: prev: {
-                  passthru = removeAttrs (prev.passthru or { }) [
-                    "__structuredAttrs"
-                    "strictDeps"
-                  ];
-                }
-              )
-            else if pkgs.lib.isDerivation value then
-              throw "${name} derivation is missing overrideAttrs"
-            else
-              value;
+          cleanPackage = overrideValue (
+            _: prev: {
+              passthru = removeAttrs (prev.passthru or { }) [
+                "__structuredAttrs"
+                "strictDeps"
+              ];
+            }
+          );
         in
         {
           AttributeSet = {
--- a/src/eval.nix
+++ b/src/eval.nix
@@ -57,11 +57,30 @@ let
       if !builtins.isAttrs value then
         { NonAttributeSet = null; }
       else
+        let
+          # Disallows people getting around actually setting `strictDeps`
+          # and `__structuredAttrs` by doing something like:
+          # {
+          #   passthru.strictDeps = true;
+          # }
+          # or
+          # package = package-final // {
+          #   __structuredAttrs = true;
+          # };
+          cleanPackage = value.overrideAttrs (
+            _: prev: {
+              passthru = removeAttrs (prev.passthru or { }) [
+                "__structuredAttrs"
+                "strictDeps"
+              ];
+            }
+          );
+        in
         {
           AttributeSet = {
             is_derivation = pkgs.lib.isDerivation value;
-            strict_deps = value.strictDeps or false;
-            structured_attrs = value.__structuredAttrs or false;
+            strict_deps = cleanPackage.strictDeps or false;
+            structured_attrs = cleanPackage.__structuredAttrs or false;
             definition_variant =
               if !value ? _callPackageVariant then
                 { ManualDefinition.is_semantic_call_package = false; }
--- a/src/eval.nix
+++ b/src/eval.nix
@@ -67,14 +67,20 @@ let
           # package = package-final // {
           #   __structuredAttrs = true;
           # };
-          cleanPackage = value.overrideAttrs (
-            _: prev: {
-              passthru = removeAttrs (prev.passthru or { }) [
-                "__structuredAttrs"
-                "strictDeps"
-              ];
-            }
-          );
+          cleanPackage =
+            if value ? overrideAttrs then
+              value.overrideAttrs (
+                _: prev: {
+                  passthru = removeAttrs (prev.passthru or { }) [
+                    "__structuredAttrs"
+                    "strictDeps"
+                  ];
+                }
+              )
+            else if pkgs.lib.isDerivation value then
+              throw "${name} derivation is missing overrideAttrs"
+            else
+              value;
         in
         {
           AttributeSet = {
--- a/.changeset/ignore_ineffective_package_attrs_in_strictdeps_structuredattrs_checks.md
+++ b/.changeset/ignore_ineffective_package_attrs_in_strictdeps_structuredattrs_checks.md
@@ -1,0 +1,9 @@
+---
+default: minor
+---
+
+# Ignore ineffective package attrs in `strictDeps` & `__structuredAttrs` checks
+
+Previously, setting e.g. `passthru.__structuredAttrs` would satisfy our checks, even though this is not effective to actually enable `__structuredAttrs` in the derivation.
+
+Detection of these features now more closely aligns with the actual derivation's use of them.
--- a/tests/mock-nixpkgs.nix
+++ b/tests/mock-nixpkgs.nix
@@ -41,11 +41,13 @@ let
     # By using `mapAttrs`, `builtins.unsafeGetAttrPos` just returns `null`,
     # which then doesn't trigger this check
     // lib.mapAttrs (name: value: value) {
-      someDrv = {
-        type = "derivation";
-        strictDeps = true;
-        __structuredAttrs = true;
-      };
+      someDrv = (
+        lib.makeExtensibleWithCustomName "overrideAttrs" (finalAttrs: {
+          type = "derivation";
+          strictDeps = true;
+          __structuredAttrs = true;
+        })
+      );
     };
 
   baseDirectory = root + "/pkgs/by-name";
--- a/tests/mock-nixpkgs.nix
+++ b/tests/mock-nixpkgs.nix
@@ -33,6 +33,15 @@ let
       newScope = extra: lib.callPackageWith (self // extra);
       callPackage = self.newScope { };
       callPackages = lib.callPackagesWith self;
+      mkFakeDrv =
+        args:
+        lib.makeExtensibleWithCustomName "overrideAttrs" (
+          finalAttrs:
+          {
+            type = "derivation";
+          }
+          // args finalAttrs
+        );
     }
     # This mapAttrs is a very hacky workaround necessary because for all attributes defined in Nixpkgs,
     # the files that define them are verified to be within Nixpkgs.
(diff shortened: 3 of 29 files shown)


src/eval.nix
```diff
@@ -60,8 +60,8 @@ let
         {
```
MattSturgeon, line 83: How about if we did something like:
```nix
let
  cleanPackage = value.overrideAttrs (finalAttrs: prevAttrs: {
    passthru = removeAttrs (prevAttrs.passthru or { }) [
      "__structuredAttrs"
      "strictDeps"
    ];
  });
in
```

Then we can reliably check `cleanPackage.__structuredAttrs or false` etc as the "workaround hacks"[^1] are discarded by overrideAttrs.


[^1]: `passthru` and `//`

RossSmyth: That works for the first case, if they define `passthru.strictDeps = true;`, but does not change the second case of 
```nix
package = package-final // {
  strictDeps = true;
};
```

mdaniels5757: One advantage of this is that `drvAttrs` can be set via `passthru` (even though it [normally is set by nix](https://github.com/NixOS/nix/blob/2c25a9aa8c561577922ea158eb372a0e5e832d74/src/libexpr/primops/derivation.nix#L29)):

```
nix-repl> (patch.overrideAttrs { passthru.drvAttrs = { fake = 1; }; }).drvAttrs
{ fake = "1"; }
```

mdaniels5757: Just saw your comment. Maybe we should have both then??

MattSturgeon: > That works for the first case, if they define `passthru.strictDeps = true;`, but does not change the second case of

It should work for both cases; `overrideAttrs` creates a new attrset, it has no knowledge of the `// { strictDeps = true; }` in your example.

<details><summary>Repl example</summary>
<p>

```nix
nix-repl> :l <nixpkgs>
Added 27224 variables.

nix-repl> hello.foo
error: attribute 'foo' missing

nix-repl> hello2 = hello // { foo = "foo"; }

nix-repl> hello2.foo
"foo"

nix-repl> hello3 = hello2.overrideAttrs {}

nix-repl> hello3.foo
error: attribute 'foo' missing
```

</p>
</details> 

(this behaviour is the entire reason `passthru` exists in the first place 😉 )

RossSmyth: Ah good point

RossSmyth: Ok next problem is getting `mock-nixpkgs` setup with `overrideAttrs`. I'm floundering a bit. I've got this so far: 
```patch
diff --git a/tests/mock-nixpkgs.nix b/tests/mock-nixpkgs.nix
index 35f246fec4..9db10d5536 100644
--- a/tests/mock-nixpkgs.nix
+++ b/tests/mock-nixpkgs.nix
@@ -41,11 +41,13 @@
     # By using `mapAttrs`, `builtins.unsafeGetAttrPos` just returns `null`,
     # which then doesn't trigger this check
     // lib.mapAttrs (name: value: value) {
-      someDrv = {
-        type = "derivation";
-        strictDeps = true;
-        __structuredAttrs = true;
-      };
+      someDrv = (
+        lib.makeExtensibleWithCustomName "overrideAttrs" (finalAttrs: {
+          type = "derivation";
+          strictDeps = true;
+          __structuredAttrs = true;
+        })
+      );
     };
 
   baseDirectory = root + "/pkgs/by-name";
```
Looking at how https://github.com/NixOS/nixpkgs/blob/master/pkgs/stdenv/generic/make-derivation.nix#js-repo-pjax-container is structured, I think I'm supposed to be binding this somewhere and then inheriting it. But I'm not sure where.

MattSturgeon: I don't think `mock-nixpkgs` is relevant here. This code seems to be evaluating a _real_ nixpkgs, with an overlay that modifies `callPackage` to inject some metadata.

I'd do:
```nix
{
  AttributeSet =
    let
      # Just in case this isn't a mkDerivation-produced derivation
      overrideValue = value.overrideAttrs or (_: value);
      # Discard any ineffective attrs passed via `passthru` or `//`
      # We only care if the derivation has __structuredAttrs, not its passthru
      cleanValue = overrideValue (finalAttrs: prevAttrs: {
        ${if prevAttrs ? passthru then "passthru" else null} = removeAttrs prevAttrs.passthru [
          "strictDeps"
          "__structuredAttrs"
        ];
      });
    in
    {
      is_derivation = pkgs.lib.isDerivation value;
      strict_deps = cleanValue.strictDeps or false;
      structured_attrs = cleanValue.__structuredAttrs or false;
      definition_variant =
	if !value ? _callPackageVariant then
	  { ManualDefinition.is_semantic_call_package = false; }
	else
	  value._callPackageVariant;
    };
};
```

RossSmyth: mock-nixpkgs is how the tests run, a real nixpkgs is not evaled

RossSmyth: For context, this is for making the tests work. They are not real nixpkgs derivations thus they must pretend to be

MattSturgeon: Your `makeExtensibleWithCustomName`-based mock already looks correct.

It'll only work if we pass full overlay-style extension functions to `overrideAttrs` (of course). The one-arg and plain-attrs overrides would require a more complex mock, closer to the actual overrideAttrs impl. Maybe using `lib.toExtension`? But I don't think the test fixture needs to support that usage.

RossSmyth: This my work so far. Tests do not eval.

MattSturgeon: > This my work so far. Tests do not eval.

I pushed up a fix; not all attrs are derivations with an `overrideAttrs`, that's why in https://github.com/NixOS/nixpkgs-vet/pull/265#discussion_r3417823215 I had `overrideValue = value.overrideAttrs or (_: value);`.

In the fix I pushed up, I included an throw for `value ? overrideAttrs` != `isDerivation value`. Throwing probably isn't the right way to handle that and may trip up on real-world examples not covered by the test suite.

One test still fails, but the failure looks correct to me (i.e. the test itself is wrong):

You have
```nix
{ someDrv }: someDrv // { strictDeps = false; }
```

expecting:
```
- Attribute `pkgs.foo` is a new package with `strictDeps` unset or set to `false`.
  Please enable `strictDeps = true;` in pkgs/by-name/fo/foo/package.nix.
```

However, `someDrv` does have `strictDeps = true` and the `// { strictDeps = false; }` in your test fixture has no real effect.

MattSturgeon: > One test still fails

Turned out it was more than one, but `cargo test` stopped at the first failure. I've pushed up fixes for the tests, though.

src/eval.nix
```diff
@@ -57,11 +57,36 @@ let
...
+              value.overrideAttrs (
+                _: prev: {
+                  passthru = removeAttrs (prev.passthru or { }) [
+                    "__structuredAttrs"
+                    "strictDeps"
+                  ];
+                }
+              )
+            else if pkgs.lib.isDerivation value then
+              throw "${name} derivation is missing overrideAttrs"
```
MattSturgeon: We want to handle this case, but a throw is probably incorrect. Ideas?
Are there any _real_ Nixpkgs attrs that'd hit this "`isDerivation` but no `overrideAttrs`" edge case?

RossSmyth: I don't think so. There are no usages of `makeExtensibleWithCustomName` in nixpkgs, so no one is making their own `overrideAttrs`. People just use `makeExtensible`. So I think it would be safe to ignore that case.

RossSmyth:
Thanks for the fix-ups.

src/eval.nix
MattSturgeon: > so no one is making their own `overrideAttrs`

That's not what's relevant here. This if-branch is for derivations that _don't_ have an `overrideAttrs` member.

In nixpkgs, `overrideAttrs` is added by `stdenv.mkDerivation`. While `isDerivation` is true for any attrset with `type = "derivation"`.

`builtins.derivation` won't add `overrideAttrs`, neither will a fake derivation like `{ type = "derivation"; }`, nor [`lib.toDerivation`](https://noogle.dev/f/lib/toDerivation/), etc.

<details><summary><code>builtins.derivation</code></summary>
<p>

```
nix-repl> d = builtins.derivation { name = "foo"; builder = "bash"; system = builtins.currentSystem; }

nix-repl> d ? overrideAttrs
false
```

</p>
</details> 

I guess we need to add a test case for a derivation without `overrideAttrs` too...

MattSturgeon: Looks like this _would_ affect real attrs in Nixpkgs, mostly `stdenv`s:

<details><summary>repl</summary>
<p>

```
nix-repl> lib.attrNames (lib.filterAttrs (_: p: (builtins.tryEval p).success && lib.isDerivation p && !p?overrideAttrs) pkgs)
[
  "aroccStdenv"
  "ccacheStdenv"
  "clangMultiStdenv"
  "clangStdenv"
  "clangStdenvNoLibs"
  "distccStdenv"
  "emscriptenStdenv"
  "fastStdenv"
  "gcc13Stdenv"
  "gcc14Stdenv"
  "gcc15Stdenv"
  "gcc16Stdenv"
  "gccCrossLibcStdenv"
  "gccMultiStdenv"
  "gccStdenv"
  "gccStdenvNoLibs"
  "gnuStdenv"
  "intelLlvmStdenv"
  "libcxxStdenv"
  "multiStdenv"
  "stdenv"
  "stdenvNoCC"
  "stdenvNoLibc"
  "stdenvNoLibs"
  "stdenv_32bit"
  "zigStdenv"
]

nix-repl> pkgs.stdenv ? overrideAttrs
false

nix-repl> lib.isDerivation pkgs.stdenv
true
```

</p>
</details>

MattSturgeon: I pushed something up, LMK what you think.

RossSmyth: Ah, sorry, I'm still waking up. I'm not aware of any raw `builtin.derivation` usages in nixpkgs outside of stdenv (obviously) and some tests. 

Doing some testing, if you do make a raw derivation, `overrideDerivation` is still added, but I don't think that really solves anything.

RossSmyth:
Your solution looks good

RossSmyth:
I am a bit curious to know if there are any packages that are used without stdenv. I guess the bootstrap packages are.

MattSturgeon (approved):
LGTM, but I authored some bits, so would be nice to get a second review.

src/eval.nix
```diff
@@ -57,11 +57,34 @@ let
       if !builtins.isAttrs value then
         { NonAttributeSet = null; }
       else
+        let
+          # Not all attrs (or even all derivations) have an `overrideAttrs`,
+          # only those constructed by something based on `stdenv.mkDerivation`
+          overrideValue = value.overrideAttrs or (_: value);
+
+          # Disallows people getting around actually settings `strictDeps`
```
mdaniels5757: :)
```suggestion
          # Disallows people getting around actually setting `strictDeps`
```

mdaniels5757 (approved):
Diff LGTM, ran locally

src/eval.nix
RossSmyth: Good catch. I'll push a fix shortly.

PR #276, opened 2026-07-07
Prep for multiple by-name directory work

mdaniels5757:
I figured it'd be easier to split these out. It will be easier to review these commit-by-commit.

- **refactor(problems): derive Debug**
- **test: use `fixtures` crate, prepare for multiple by-name directories**
- **refactor(problems): don't hardcode `pkgs.` prefix in attrnames**


philiptaron (approved):
That's a lot of files

philiptaron:
I had left `Debug` off for essentially the same reasons you're turning it on (no production use) but I have no issue with flipping the switch the other way.

PR #291, opened 2026-08-04
ci: set concurrency.cancel-in-progress correctly

mdaniels5757:
The comment was correct, the code was not.

E.g. https://github.com/NixOS/nixpkgs-vet/actions/runs/30952130007/job/92136500364

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ concurrency:
   # We don't want to cancel in progress workflows for pushes to the main branch,
   # because it could be in the process of doing a release.
   # Better wait until it's done
-  cancel-in-progress: ${{ github.event_name != 'pull' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -17,7 +17,7 @@ concurrency:
   # We don't want to cancel in progress workflows for pushes to the main branch,
   # because it could be in the process of doing a release.
   # Better wait until it's done
-  cancel-in-progress: ${{ github.event_name != 'pull' }}
```
mdaniels5757, line 20: "Why would this not have been noticed to not cancel on some PR before?", you may ask. The answer: `pull` is not _ever_ the `event_name` used (it's `pull_request`).

philiptaron (approved):
sigh

